PR #3, opened 2019-08-03
Rhel8

johnnykeats:
RHEL 8 support.
NIC role is broken due to known bug, currently investigated in Ansible team.


oxedions:
I can't merge it. And it's my fault, sorry. I changed too much things in the mean time.

I propose we drop this branch, and I will merge all manually using Meld.

Also, these packages are a pain, I will stop adding so much sources. And we will clean old commits.

johnnykeats:
You will never change dude ( ^ ᗜ ^ )
Ok, let's do it this way. And I propose to create a specific repository to store packages if we really need them.

oxedions:
Ok !
I will do it asap :-)
And ok to store packages outside. We keep spec files or patchs in the main repo.

PR #11, opened 2019-09-14
Adapting slurm to handle both packaging. Allow OpenHPC slurm.

oxedions:
Hello JK.

Someone asked for OpenHPC compatibility.

All HPC scientific packages from OpenHPC are autonomous, so ok for them. Only issue I see is with slurm version.

Then, BlueBanquise will be able to replace provisioning of OpenHPC, but be compatible with it's rpms.

Is it ok for you this way ?

Ox

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,10 @@ Currently supported OS are:
 
 Debian 9 and 10 are dropped for now because too much issues with the netboot part.
 
-Ansible **>= 2.8.2 is mandatory** for BlueBanquise to run properly.
+Ansible >= 2.8.2 is mandatory for BlueBanquise to run properly.
+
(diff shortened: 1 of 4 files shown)


johnnykeats:
Ok for me. But we definitely need to build our packages also.

I found Nehalem, Sandy, and Haswell. Maybe someone with Skylake, not sure. Nobody with AMD :confused:
Also Cortex-A53 for aarch64 (PI3). I think we can rely on Nehalem for generic packages, no need to handle previous generations, gap is to big.

oxedions:
Merged.

Yes, http server should be up soon also.

I agree with Nehalem for generics, gap is to big before these.

Ox

PR #21, opened 2019-10-23
bugfix in conman role

bouriquet15:
fix missing "}" in conman role
update from with_items to loop
add notify (not fully implemented yet)


oxedions:
Nice handler. I definitely need to add more of them.

Thanks !

Oxedions

PR #22, opened 2019-10-23
Bugfix/ipxe template

bouriquet15:
Add temp variables :
- Reduce hostvars calls
- Improve readability of kernel line
- Should be easier to maintain

Bugfix : internal var "equipment_naming" were not used here

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,6 +1,6 @@
 # Roadmap
 
-Current status: 1.1.0 RC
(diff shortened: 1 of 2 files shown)


oxedions:
Hello @bouriquet15 

Many thanks for this bug fix and the improvements!
Code is definitely clearer now in ipxe templates files.

I rebased to 1.1.1 branch.

Oxedions

PR #40, opened 2019-11-29
Fix iPXE os deployment is not working on RHEL 7 (and probably on Centos 7)

gavillom:


oxedions:
Dear @gavillom ,

Many thanks for this PR :-)

This is indeed a bug, and it impact also Centos and RHEL 8 versions.

I gladly accept this PR, and I will fix on my side the 8 versions.

Ben

PR #44, opened 2019-12-03
Add external repositories

bouriquet:
Add support for external repositories (CentOS 8 / RedHat 8 only)
Minor:
* Concatenate BaseOS / AppStream


oxedions:
I will merge, but edit later as I find using system_repositories is not more logical. Internal and external are all "system".

Many thanks for this PR :-)

Ben

PR #46, opened 2019-12-12
advanced_dhcp_server role fails with opt61

hmeAtos:
Hi,
When I write an inventory file that describes the options 61, I have issue:
```
failed: [mngt0-1] (item=ice1-1) => changed=false
  ansible_loop_var: item
  item: ice1-1
  msg: 'AnsibleUndefinedVariable: ''dict object'' has no attribute ''dhcp'''
```
I have fixed this issue like this.

Kind regards


oxedions:
Dear @hmeAtos 

You are right, I inverted both values.

Many thanks for this fix :-)
I will report it into 1.2.0 branch also.

Best regards

Ben

PR #55, opened 2019-12-17
Fix: rename of system_repositories in tasks

actatux:
Commit d60214be replaced system_repositories with repositories in the
examples and tasks for CentOS 8. Update the repositories_client tasks
for other OS as well.


oxedions:
Dear @actatux ,

Thank you for this PR.
This is indeed a bug.
I shall also correct the when statement, not filtering all issues (should check iterable also).

With my best regards

Ben

PR #56, opened 2019-12-17
config: use /etc/bluebanquise as default directory

actatux:
Avoid using the default /etc/ansible directory for BlueBanquise. Some
users may need additional Ansible playbooks or install other tools that
may override some files in /etc/ansible.

As stated in the configuration file, ansible will read ANSIBLE_CONFIG,
ansible.cfg in the current working directory, .ansible.cfg in the home
directory or /etc/ansible/ansible.cfg, whichever it finds first.

We might recommand to change the current workdir to /etc/bluebanquise,
or set the ANSIBLE_CONFIG. If a user does not plan to use ansible for
any other purposes, it is still possible to create a symlink
/etc/ansible/ansible.cfg -> /etc/bluebanquise/ansible.cfg.

Also add the pxe directory into .gitignore.


oxedions:
Dear @actatux 

Thank you for this PR.

As it changes many aspects of the stack (mostly documentation, but also some "philosophy"), I need to check with the users quorum their thought about this. :-)

It may take some time.

Ox

oxedions:
As discussed, merging.

Ben

PR #57, opened 2019-12-18
Feat: make yum repositories more configurable

actatux:
Allow to modify some settings for the repo, be it a system repository or
an external repository.

List of settings: name, baseurl, enabled, gpgcheck, gpgkey and proxy.

System repositories, internal to the cluster, can use the same
configuration syntax than previously:

  repositories:
    - bluebanquise
    - os

It can be expanded to change settings for a repo with a dict, using
YAML's flow styles or block styles:

  repositories:
    - bluebanquise
    - { name: 'os', gpgcheck: 1 }  # Flow style
    - name: epel                   # Block style
      gpgcheck: 1

Additional repositories can either be added to the repositories.yml or
external.yml configuration files, to logically segregate internal and
external repositories.

  external_repositories:
    - name: epel
      baseurl: 'https://dl.fedoraproject.org/pub/epel/$releasever/$basearch/'
      proxy: 'https://proxy:8080'

### DIFF
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/external.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/external.yml
@@ -22,7 +22,8 @@ external_hosts:
 # Be aware that, at this level, theses repositories will be exported on all hosts without
 # any architecture or distribution check.
 
-external_repositories:
+#external_repositories:
 #  - name: epel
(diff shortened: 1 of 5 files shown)


oxedions:
Dear @actatux ,

Many thanks for this PR.
This is ok for me, perfect.

We will need to add some extra documentation, but at least all features needed are now available :-)

Merging.

Ben

PR #61, opened 2019-12-21
Fix: replace symlinks from simple_cluster examples

actatux:
Commit 4c24c8f7 introduced two symbolic links in the simple_cluster
inventory, with relative paths to the multi_icebergs_cluster inventory.

The "Configure BlueBanquise" documentation provides commands to copy the
simple_cluster inventory as the new inventory. Doing so breaks the
relative symlinks and makes the inventory unusable.

This commit fixes this by removing the symlinks and copying the
directories of the multi_icebergs_cluster inventory to the
simple_cluster inventory.


oxedions:
Dear @actatux 

Thank you for this PR.
I agree. This means I will have to maintain both versions when doing modifications, but this is ok as multiple persons faced this issue.

Ben

Issue #60, opened 2019-12-18
Need to have 2 disctinct services for tftp and http in pxe chain boot


jpm38:
Bluebanquise stack uses the same IP address in pxe chain boot for tftp service and http service: The pxe_ip variable in network definition.
networks:
  ice1-1:
    subnet: 10.10.0.0
    prefix: 16
    netmask: 255.255.0.0
    broadcast: 10.10.255.255
    dhcp_unknown_range: 10.10.254.1 10.10.254.254
    gateway: 10.10.255.254
    is_in_dhcp: true
    is_in_dns: true
    services_ip:
      pxe_ip: 10.10.0.1

This is working fine when services are running on the same host/VM/container and so exposing the same IP address for both services.
When deploying services like dhcp, tftp, http, dns etc. inside containers running on various bare metal nodes, orchestrated bu Kubernetes, we have a service IP address per service. For that reason, we need to have a pxe chain boot solution for which we can configure a tftp_ip address and a pxe_ip address for respectively access to TFTP server and HTTP server.

oxedions:
Dear @jpm38 ,

I understand your need.

The PXE chain was designed to be as simple as possible for debugging, as I found PXE part is always the trickiest part.

However, having the ability to split these 2 ip is understandable.

I can propose you the following way, only for advanced_dhcp_server role:

If in network ice1-1 for example, the service pxe_http_ip is set (assuming here 10.10.77.1), then dhcp template will add in the ice1-1 subnet:

`option root-path "10.10.77.1";`

And then, in the embed script of the rom, there would be:

`isset ${root-path} && set next-server ${root-path}`

This way, next-server ip would be override by this new ip.

I know root-path was not design for an ip, but for an iSCSI target. However, the probability someone use the BlueBanquise iPXE rom for iSCSI targets is very very low...

Would this be a good solution for you ?

Ox

jpm38:
Dear @oxedions 

Your proposal looks good to me and fit exactly our needs.
Can I hope to have this feature in 1.2 ?
Many thanks

JPM

oxedions:
Dear @jpm38 ,

@actatux PR provided a solution.

I will rebuild new roms and you will be able to benefit from it, in 1.2.0 version :-)

Ben

PR #62, opened 2019-12-22
Feature: alternate next-server for HTTP/iPXE

actatux:
By default, the pxe_ip defines the next-server to use for both the TFTP
and HTTP/iPXE services which muse be hosted on the same server.

It is now possible to use a different next-server for HTTP/iPXE by
defining a new service parameter pxe_http_ip, thus the HTTP server for
ipxe scripts can be run on a separate server than the TFTP service.

The PXE chain remains unchanged: the first next-server obtained at boot
time is pxe_ip and it allows to load the iPXE bluebanquise ROM through
TFTP. This ROM will send a new request to the DHCP to get the
next-server to use for the HTTP/iPXE service. If pxe_http_ip is set,
then this is the IP address that will be used as next-server. Otherwise,
the default will be the pxe_ip.

Solve #60.


actatux:
Depends on bluebanquise-infrastructure#1

oxedions:
Dear @actatux ,

Thank you for this PR.
To be combined with https://github.com/oxedions/bluebanquise-infrastructure/pull/1 as a reference.

I will build new set of iPXE Roms.

Ben

PR #66, opened 2020-01-02
Update slurm.conf.j2

sobrase:
testé et fonctionnel


oxedions:
Dear @sobrase 

Unfortunately, this loop.first will not work, as it will rely on the nested loop, while the first level loop is the one to consider. However, adding a simple "," even if there is no host before is ok for slurm, so the patched version of the 1.2.0 branch is I think ok now.

Do you agree on this ?

Ben

oxedions:
Discussed internally with @sobrase, we agreed to keep current file as it is.

PR #67, opened 2020-01-02
Update centos_7.yml

sobrase:
ne marche pas avec file d'où le shell

### DIFF
--- a/roles/core/repositories_client/tasks/centos_7.yml
+++ b/roles/core/repositories_client/tasks/centos_7.yml
@@ -7,3 +7,8 @@
     baseurl: "http://{{networks[j2_node_main_network]['services_ip']['repository_ip']}}/repositories/{{equipment_profile['operating_system']['distribution']}}/{{equipment_profile['operating_system']['distribution_version']}}/{{equipment_profile['hardware']['cpu']['architecture']}}/{{item}}"
     gpgcheck: no
   with_items: "{{ repositories }}"
+  
+- name: Remove repository
+  shell: rm -rf /etc/yum.repos.d/centOS*
+  args:
+    warn: false


roles/core/repositories_client/tasks/centos_7.yml
```diff
@@ -7,3 +7,8 @@
     baseurl: "http://{{networks[j2_node_main_network]['services_ip']['repository_ip']}}/repositories/{{equipment_profile['operating_system']['distribution']}}/{{equipment_profile['operating_system']['distribution_version']}}/{{equipment_profile['hardware']['cpu']['architecture']}}/{{item}}"
     gpgcheck: no
   with_items: "{{ repositories }}"
+  
+- name: Remove repository
```
actatux: Remove any previous repository *before* creating new ones. If a user create a repo with prefix centOS, this change will remove it.

roles/core/repositories_client/tasks/centos_7.yml
```diff
@@ -7,3 +7,8 @@
...
+  shell: rm -rf /etc/yum.repos.d/centOS*
```
actatux: s/centOS/CentOS-/

oxedions:
Dear @sobrase ,

As for the nfs_server, can you add a variable, false by default, that would allow to activate/deactivate this feature ?

For example, on the Tumulus cluster of the Algoric HPC Center, as the cluster is small, we do not sync update repos from Centos on the cluster, and so we need these CentOS repos to stay on computes nodes :-)

Ben

actatux:
> For example, on the Tumulus cluster of the Algoric HPC Center, as the cluster is small, we do not sync update repos from Centos on the cluster, and so we need these CentOS repos to stay on computes nodes :-)

You should definitely use the new external_repositories parameter. :-)

I would prefer to not delete the existing files, but it is definitely an issue when a node cannot access the CentOS mirrors. I'm wondering if we could achieve something less intrusive with the ini_file module to disable the repos (enabled=0).

I will give a try this week.

oxedions:
Dear @actatux 

Ok, thanks :-)

Ben

actatux:
With CentOS 8.1, the distro repositories are now installed by package `centos-repos`, thus we can ensure this package is absent if we don't want to use the distro repositories. See https://git.centos.org/rpms/centos-release/c/26a0d73cedb9a2bfbcc38459344111fea231c577

actatux:
Upstream dev eventually harden the `centos-repos` dependency... https://git.centos.org/rpms/centos-release/c/21470e447157e90e0fd6b63047b2cc7040884748?branch=c8

actatux:
@sobrase @oxedions please review #88 which is an alternative to this PR.

oxedions:
Closing, see #88

PR #68, opened 2020-01-02
Update main.yml

sobrase:


roles/core/nfs_server/tasks/main.yml
```diff
@@ -20,6 +20,12 @@
 
 - meta: flush_handlers
 
+- name: Create directory                                                                                                                                                                                             file:
+    path: "{{ nfs[item]['export'] }}"
```
actatux: I guess it should be nfs[item]['mount'] which is the path the NFS server should export?

actatux:
Also, could you describe the change in your commit title and message? There is at least one main.yml per role, so we can't know which role is modified or why it is modified with the current commit message. It would be very helpful when looking at the git log.

eg.: **Feat: Create the NFS directories to export**

Thanks :)

oxedions:
Dear @sobrase ,

I honestly would prefer to skip directory creation on the nfs_server part. While I find it relevant in nfs_client part, I think it can be difficult to handle, or dangerous on the nfs_server part. At least, there must be a variable, false by default, to activate/deactivate this feature :-)

Ben

oxedions:
No activity, and still don't think creating directories automatically is a good idea on server side.

Closing.

Issue #59, opened 2019-12-18
Need script and directory tags in pxe_stack role


jpm38:
When deploying services like dhcp, tftp, http, dns etc. inside containers running on various bare metal nodes, we used the ansible "chroot" connector to generate the configuration files on a shared file system mounted on all nodes. 
We want to use the standard roles (dhcp, dns, pxe_stack ...) for that purpose, but we do NOT want to start the services on the bare metal management node. So we use the template tag , but we need also for pxe_stack role to create the output directories and to copy the bootswitch cgi script. So, a directory and a script tag are needed for that.

oxedions:
Dear @jpm38 ,

Thank you for this proposal. I agree it would be a good feature.

Could you propose a PR based on branch 1.2.0 ?

Many thanks !

Ox

jpm38:
Dear @oxedions 

Sounds goot to me.

JPM

actatux:
Could you achieve the same behavior (don't start services) with --skip-tags "service"?

jpm38:
I've tested your proposal adding also the skip of "package":
--skip-tags package,service

and it does the trick.

thanks

actatux:
Can we close then?

jpm38:
yes, sure

oxedions:
Closing.

PR #70, opened 2020-01-07
Need script and directory tags in pxe_stack role

jpm38:
When deploying services like dhcp, tftp, http, dns etc. inside containers running on various bare metal nodes, we used the ansible "chroot" connector to generate the configuration files on a shared file system mounted on all nodes.
We want to use the standard roles (dhcp, dns, pxe_stack ...) for that purpose, but we do NOT want to start the services on the bare metal management node. So we use the template tag , but we need also for pxe_stack role to create the output directories and to copy the bootswitch cgi script. So, a directory and a script tag are needed for that.

### DIFF
--- a/roles/core/pxe_stack/tasks/main.yml
+++ b/roles/core/pxe_stack/tasks/main.yml
@@ -19,6 +19,8 @@
    - /var/www/cgi-bin
    - /etc/bluebanquise
    - /etc/bluebanquise/pxe
+  tags:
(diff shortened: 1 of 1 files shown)


actatux:
Is this PR still valid after discussion in #59?

jpm38:
I think notm qs the solution you proposed does the trick.

oxedions:
Closing.

PR #72, opened 2020-01-07
consistency of tag names in roles

hmeAtos:
Hi,
We noticed that some tag names were not consistent between the different roles.
I propose this patch.
Best regards


hmeAtos:
we noticed that some tag names were not consistent between the different roles.

PR #88, opened 2020-01-27
feat: Disable CentOS distro repositories

actatux:
This feature allows to disable the CentOS repositories enabled by
default after the installation of the OS. It can be enabled or disabled
with the boolean parameter remove_distro_repositories.

If the parameter is not defined, the repositories are untouched to keep
current behavior. In the examples, the parameter defaults to True
because it is good practice to set up a local mirror when we have a lot
of systems.


oxedions:
Hello @actatux 

My apologies for being so late. Back online.

Perfect for me, I took time to check your code and seems ok for me.

So this replace https://github.com/oxedions/bluebanquise/pull/67

Merging.

Ben

PR #89, opened 2020-01-28
fix: debug slurm role

actatux:


roles/addons/slurm/tasks/main.yml
```diff
@@ -75,7 +75,7 @@
     name: slurmctld
     state: started
     enabled: no         # Dangerous to enable slurm controller...
-  when: slurm_profile == 'controler'
+  when: slurm_profile == 'controller'
```
actatux: To avoid breaking existing configuration, we can consider `when: slurm_profile == 'controller' or slurm_profile == 'controler'` with a deprecation notice.

oxedions:
Dear @actatux ,

Thank you for this FIX.

For what I know, no one reported usage of this feature for now, so we can merge without impacting users.

Ben

actatux:
This is a breaking change for slurm controller.

PR #90, opened 2020-01-30
fix: debug diskless role

actatux:


oxedions:
Dear @actatux 

Checked, good for me.
Many thanks !!

I let the issue open, so that I remember to revamp the whole role asap.

Merging

Ben

PR #92, opened 2020-01-30
fix: [pxe_stack] remove recursive setting of perms

actatux:
The recursive task was changing mod of the files in the directory, while
following task are setting different mod for the same files after.

With this change, a new run of the pxe_stack role will not report change
if the configuration of the target is up-to-date.


oxedions:
Dear @actatux 

Very nice ! Thanks for this :-)

merging

Ben

PR #97, opened 2020-02-07
Fix: Solve issue for j2 network file

oxedions:
Issue when an host do not have an administration network connected (iceX-Y)


oxedions:
Dear @actatux 

I agree. I was wondering what would cost the most: storing this into a variable (so using a set), or computing this 2 times...

I never designed this for hosts that would be member of any administration networks :-(

I am really thinking of forcing this as a requirement, and so remove this patch...

oxedions:
Not relevant for now. Closing.

PR #104, opened 2020-02-17
[1.1.2] Trailing whitespace in dhcpd.subnet.conf.j2

hmeAtos:
Hello,
When we define several networks (ice1-2, ice1-3, ...), the `dhcpd.subnet.conf.j2` template generate many **trailing whitespace and blank line**s in DHCP configuration files.
This is not a problem to run the DHCP service, but the configuration files generate is too big and not very clean.

For fix this issue, please merge my patch.

Regards,

### DIFF
--- a/roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
+++ b/roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
@@ -32,14 +32,15 @@
       {% set ipxe_rom_filename = (ipxe_architecture + '/standard_' + ipxe_driver + 'ipxe.efi') %}
     {% else %}  {# Assume everything else is pcbios #}
       {% set ipxe_rom_filename = (ipxe_architecture + '/standard_undionly.kpxe') %}
-    {% endif %}
+    {% endif -%}
 
-    {% for nic, nic_args in hostvars[host]['network_interfaces'].items() %}
+    {%- for nic, nic_args in hostvars[host]['network_interfaces'].items() %}
       {% if (nic_args.network is defined and not none) and (nic_args.network == item) and (nic_args.ip4 is defined and not none) %}
+
         {% if nic_args.mac is defined and not none %}
-  host {{host}} { 
+  host {{host}} {
     option host-name "{{host}}";
-    hardware ethernet {{nic_args.mac}}; 
+    hardware ethernet {{nic_args.mac}};
     fixed-address {{nic_args.ip4}};
     filename "{{ipxe_rom_filename}}";
   }
@@ -71,11 +72,12 @@
   }
         {% endif %}
       {% endif %}
-    {% endfor %}
+    {% endfor -%}
 
-    {% if hostvars[host]['bmc'] is defined %}
+    {%- if hostvars[host]['bmc'] is defined %}
       {% set bmc_args = hostvars[host]['bmc'] %}
       {% if (bmc_args.network is defined and not none) and (bmc_args.network == item) and (bmc_args.name is defined and not none) and (bmc_args.ip4 is defined and not none) %}
+
         {% if bmc_args.mac is defined and not none %}
     host {{bmc_args.name}} {
       option host-name "{{bmc_args.name}}";
@@ -110,4 +112,3 @@
     {% endif %}
   {% endif %}
 {% endfor %}
-


roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
```diff
@@ -32,14 +32,15 @@
       {% set ipxe_rom_filename = (ipxe_architecture + '/standard_' + ipxe_driver + 'ipxe.efi') %}
     {% else %}  {# Assume everything else is pcbios #}
```
actatux: Trailing whitespaces are between the {% else %} block and the comment block. Remove the 2 spaces before each comment and the destination file should look better.

roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
```diff
@@ -32,14 +32,15 @@
...
       {% set ipxe_rom_filename = (ipxe_architecture + '/standard_undionly.kpxe') %}
-    {% endif %}
+    {% endif -%}
```
actatux: The minus sign should remove whitespace after the block, but there is no whitespace here?

roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
```diff
@@ -32,14 +32,15 @@
...
 
-    {% for nic, nic_args in hostvars[host]['network_interfaces'].items() %}
+    {%- for nic, nic_args in hostvars[host]['network_interfaces'].items() %}
```
actatux: The minus sign should remove whitespace before the block here, it is redundant with `lstrip_blocks: "True"` (1st line of the file)

roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
```diff
@@ -71,11 +72,12 @@
   }
         {% endif %}
       {% endif %}
-    {% endfor %}
+    {% endfor -%}
```
actatux: The minus sign should remove whitespace after the block, but there is no whitespace here?

roles/advanced-core/advanced_dhcp_server/templates/dhcpd.subnet.conf.j2
```diff
@@ -71,11 +72,12 @@
...
 
-    {% if hostvars[host]['bmc'] is defined %}
+    {%- if hostvars[host]['bmc'] is defined %}
```
actatux: The minus sign should remove whitespace before the block here, it is redundant with `lstrip_blocks: "True"` (1st line of the file)

bouriquet:
Hi @actatux 

Actually that's more a problem with dhcpd file being filled with many blank lines, not only spaces.

minus are here to keep the separation between the 2 pieces of code, but blank line between the two minus can just be removed, the result will be the same.

Also, lstrip doesnt strip entire blank line (and we dont expect him to do this)

hmeAtos:
Dear @actatux,
I have rebased this PR in 1.2.0 (master)
Regards

bouriquet:
Hi @hmeAtos 

You changed 600 files and added 400k lines with this rebase, not sure that what intended ? 
Looks like you rebased the entire branch, can you apply your patch on your forked master branch and submit it again ?

actatux:
Hi @bouriquet 

> minus are here to keep the separation between the 2 pieces of code,
> but blank line between the two minus can just be removed, the result will be the same.

Indeed the minus removes vertical whitespace \n. However, I do believe the blank line 36 should be removed by `trim_blocks` which defaults to yes in the template module. If not, we may dig further.

**trim_blocks**:
_Determine when newlines should be removed from blocks.
When set to yes the first newline after a block is removed (block, not variable tag!)._

Line 35 is the end of an if block.

> Also, lstrip doesnt strip entire blank line (and we dont expect him to do this)

Right, it should strip the whitespaces from the start of the line to the block. Same than the minus. If you remove these minus at lines 37 and 77, is the output different?

bouriquet:
trim_blocks is enabled here, but that's not this option purpose.

trim_blocks does not remove blank lines at all, it removes \n generated by instructions blocks.
Exemple from https://jinja.palletsprojects.com/en/2.11.x/templates/#whitespace-control :
> For example, without the trim_blocks and lstrip_blocks options, this template:
> 
> ```
> <div>
>     {% if True %}
>         yay
>     {% endif %}
> </div>
> ```
> 
> gets rendered with blank lines inside the div:
> 
> ```
> <div>
> 
>         yay
> 
> </div>
> ```

lstrip removes whitespaces before instructions blocks, allowing you to indent your jinja code

**Let's take the previous exemple :**
> 
> ```
> <div>
>     {% if True %}
>         yay
>     {% endif %}
> </div>
> ```

**Render with trim_blocks set to yes only (no lstrip)  :**
> 
> ```
> <div>
>                 yay
> </div>
> ```

_Here, the spaces before "if" instructions are kept, but newline generated by instructions blocks are removed._

**Render with lstrip set to true only (no trim_blocks) :**
> 
> ```
> <div>
>
>         yay
> 
> </div>
> ```

_Here, the indentation is correct, but we have newlines due to missing trim_blocks_


You can check at home, theses 2 options will never apply to blank lines (only affects instructions rows), minus sign does affect blank lines and any kind of formating between two minus sign.

actatux:
Interestingly when I ran ansible to generate a template with `trim_blocks: "False"` in the header, it did not add additional lines in the destination file, as it should do when I look at your example.
Anyway, I got your point, using the minus sign at the end of a block allows to remove the newline whitespace and so provide better reading of the template without impacting the rendering.
In my test, the minus sign at the beginning of the block did not change anything to the output.

bouriquet:
Remove your quotes after trim_blocks header

cat template_test.j2

```
#jinja2: trim_blocks: False

{% if true %}
  test1
{% endif %}

{% if true %}
  test2
{% endif -%}

{%- if true %}
  test3
{% endif %}
```

ansible -m debug -a "msg={{ lookup('template', 'template_test.j2') }} localhost

```
ok: [localhost] =
  msg:


         test1



         test2

         test3
```

actatux:
Indeed. I added the quotes because of https://github.com/ansible/ansible/issues/10725#issuecomment-249494236
Thanks for the tip!

oxedions:
Dear @all 

Thank you for this bugfix.

Merging.

Ben

PR #106, opened 2020-02-19
fix: add target variable to example playbooks

actatux:
Allow to use `--extra-vars target=foo` whatever the playbook.


oxedions:
Dear @actatux 

Thank you for this small but interesting change :-)

Merged.

Ben

PR #115, opened 2020-02-26
[1.2] change conman execpath

bouriquet:


oxedions:
Dear @bouriquet 

Thank you for this PR.

Can you confirm that on RHEL/Centos 7, path is also the same?

If not, we will need to ensure both distributions can be covered, and so add an if.

Best regards
Ben

bouriquet:
Hi @oxedions,

you are right, the official RH7 rpm does not provide theses files in the same path (i didn't tested it on CentOS), PR edited.

oxedions:
Many thanks.

Merging.

PR #120, opened 2020-03-05
Fix: Fixing some issues in prometheus roles

oxedions:


roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -78,8 +78,8 @@
   template:
     src: ipmi_config.yml.j2
     dest: /etc/ipmi_exporter/ipmi_config.yml
-    owner: root
-    group: root
+    owner: ipmiexporter
```
actatux: I guess this user is created by ipmi_exporter RPM?
Where can we find this RPM?

oxedions:
Dear @actatux 

Yes, it is created by rpm. I used root until now, but we need to use a standard user for such tool.

I will add this rpm into bluebanquise infrastructure, with also conman and slurm exporter. Scheduled for next week :-)

Ben

PR #121, opened 2020-03-05
Fix: Allow more main network naming patern

oxedions:
Major change, will need to play again whole deployment to ensure no issues.


oxedions:
Dear @actatux ,

I will test the fix soon, so don't bother testing it right now :-)

Once test is done, let's merge.

PR #122, opened 2020-03-05
Add configuration path variable to log roles

bouriquet:


oxedions:
Ok for me, please note that 1.1.2 is no more supported :-)

Merging !

Issue #101, opened 2020-02-14
An wrong ifcfg file is generate even no ip is set in the nic declaration


loar38:
We need to declare a nic in a new network but we can't set an IP as the IP will be mamanged by kubernetes.
I tried addind a test to skipped the ifcfg generation if no ipv4 is set in the role nic 
For example

```
- name: Set NIC configuration
  template:
    src: ifcfg.j2
    dest: /etc/sysconfig/network-scripts/ifcfg-{{item}}
    owner: root
    group: root
    mode: 0644
  with_items: "{{network_interfaces}}"
  when: (network_interfaces[item]['ip4'] is defined and not none)
```
without the test, an ifcfg file is generate but with empty parameter.

actatux:
Could you paste the generated ifcfg file without your patch?

loar38:
Here is the generated ifcfg file
[root@omn2 ~]# cat /etc/sysconfig/network-scripts/ifcfg-k8s
#### Blue Banquise file ####
## Ansible managed: modified on 2019-11-26 10:23:51 by root on hmn2

NAME=k8s
DEVICE=k8s
ONBOOT=yes
NM_MANAGED=yes
BOOTPROTO=none

TYPE=Ethernet



GATEWAY=20.50.0.2

oxedions:
Dear @loar38 

Yes, you are right, a **when** condition with presence of **ipv4** should be set in the task.
Would you accept to provide your patch as a PR?

Please consider that I made a mistake in many files, and that this:

`  when: (network_interfaces[item]['ip4'] is defined and not none)`

Is not right. You have to use instead:

`  when: (network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none)`

I made this mistake at many places of the stack (I thought it was the same, but in fact it isn't), and will correct it soon :-)

Ben

loar38:
Dear @oxedions 
Yes for sur sure, I will open 2 PR one for 1.1.2 version and one for 1.2.1 version
Using the test you propose
Laurent

oxedions:
Dear @loar38 

Did you find time to open the PR ? One for master branch will be enough :-)

Ben

oxedions:
Solved. Thanks all !
Closing.

PR #135, opened 2020-03-23
Add test for nic without ip4

loar38:
An wrong ifcfg file is generate even no ip is set in the nic declaration #101


oxedions:
Dear @loar38 

Thank you for this PR.

Merging.

PR #138, opened 2020-03-25
Fix: Add server and network_interfaces check for conman and pxe_stack…

oxedions:
… roles


actatux:
Is an equipment type **server** any node in the cluster? (service, login, compute, etc)

oxedions:
Yes. Is considered a **server** any equipment that use PXE.

oxedions:
Nop, good point. I will add it to my todolist of the documentation.
I need to improve Ansible training part, this would be the good time to do it.

PR #149, opened 2020-04-01
Use repo variables in repositories_client

actatux:
RHEL/Centos and OpenSUSE provide repo variables to get the release
version and the base architecture. This is more dynamic than relying on
the inventory.

The URL will be:
    http://repository_ip/repositories/centos/$releasever/$basearch/repo

Note that _$releasever_ in RHEL/CentOS is the major release (ie. 7 or 8),
while the previous value was the _distribution_version_ (7.6, 8.0).

Users will have to change the path of the repositories of their existing
cluster (as they would have to change the path to upgrade to the next
minor release).

If a user needs to keep minor release granularity, it is possible to overwrite
the default `$releasever` value with the inventory parameter
`repositories_releasever`.

This change might help to upgrade the systems without having to
update the inventory after each minor release.
    
See:
https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-variables
https://doc.opensuse.org/projects/libzypp/HEAD/zypp-repovars.html


oxedions:
Wouldn't that prevent users to, for example, deploy part of the cluster in RHEL 8.1 while cluster is still in in production in RHEL 8.0, to test stability? (MOFED upgrade, Lustre upgrade, Nvidia upgrades, etc).

actatux:
> Wouldn't that prevent users to, for example, deploy part of the cluster in RHEL 8.1 while cluster is still in in production in RHEL 8.0, to test stability?

Not more than preventing users to deploy a subset of nodes with 8.1 + Patchset 1 while the production is running 8.1. The RHEL 8 release cycle is [every 6 months](https://access.redhat.com/support/policy/updates/errata/#RHEL8_Life_Cycle), but Red Hat will continue to publish RHSAs and RHBAs in the meantime.
To handle such need, IMHO, one should instead use 2 states for the repositories, for example _staging_ and _production_, with different paths.

oxedions:
It took me time to compute this, and I checked with many other peoples as this implies many things.

This will bring 2 issues for me and them:

1 . We loose simple minor-release granularity. So this is a regression. BlueBanquise was designed to provide minor-release and architecture granularity. We need this at the Algoric center, and I also need it in my main work.
I have seen many issues migrating from a minor to another, and some of our users sometime ask for very specific versions while the cluster is running main production with another minor.
Best way to achieve this PR without losing this feature would be to do this:
/repositories/redhat/8/x86_64/os_erratas
/repositories/redhat/8.0/x86_64/os_erratas link to /repositories/redhat/8/x86_64/os_erratas
/repositories/redhat/8.1/x86_64/os_erratas link to /repositories/redhat/8/x86_64/os_erratas
This allows the same, while still keeping minors granularity.

2 . The idea of using "facts" (even if this is run by dnf/yum, these can be consider facts) for repositories remove ability to manually manipulate them. This is a very important behavior for all of us :-)

actatux:
> 2 . The idea of using "facts" (even if this is run by dnf/yum, these can be consider facts) for repositories remove ability to manually manipulate them. This is a very important behavior for all of us :-)

About `$basearch`, what can of manipulation could one do? Setting `aarch64` for an `x86_64` host? I may miss something. :-)

actatux:
> It took me time to compute this, and I checked with many other peoples as this implies many things.
> This will bring 2 issues for me and them:

I would be very pleased to discuss with you and them to get a better understanding. I feel dumb to not understand your concern with my sysadmin background... :-(

> 1 . We loose simple minor-release granularity. So this is a regression. BlueBanquise was designed to provide minor-release and architecture granularity.

1.a. I'm afraid most people will install their cluster with 7.4 in the path, then sync the default update stream (CentOS) or unique update stream (standard RHEL subscription) to the directory without even realizing they are syncing the latest 7 release. It means that people might believe they are running 7.4 while they are running 7.8. (yes, that's a true story).

1.b. How can you achieve architecture granularity on the software side? CPU architecture is bound to hardware. For a given host, if `basearch=x86_64` but the admin defines `equipment_profile['hardware']['cpu']['architecture']=aarch64`, he might expect some issues...

> We need this at the Algoric center, and I also need it in my main work.

You should [double check your conf](https://github.com/bluebanquise/bluebanquise/pull/67#issuecomment-571042552) before CentOS 8.2 will roll out. ;-)
It's fun you need it in your work, because in mine we don't want to manage minor releases.

> I have seen many issues migrating from a minor to another, and some of our users sometime ask for very specific versions while the cluster is running main production with another minor.

I would rather manage environments to achieve this:
`http://repository_ip/repositories/centos/$environment/$releasever/$basearch/repo`

Where `environment` can be production, staging, whatever.

Anyway, I'm open for a talk because I really don't understand what kind of granularity you are trying to achieve with minor-release and architecture.

strus38:
https://avinetworks.com/docs/18.2/how-to-tie-a-system-to-a-specific-update/

actatux:
Thank you @strus38, this might be the answer to the issue here.
This is compatible with my first commit using `$releasevar` and `$arch` in the baseurl, since it is possible to override the default value of $releasevar with the following:

```bash
[root@management1 ~]# echo '8.1.1911' > /etc/dnf/vars/releasever
```
It would need additional work to implement and document it. I'll update this PR accordingly.

actatux:
Please consider this PR.
It was tested with both CentOS 7 (7.5 images for pxe boot and 7 for packages) and CentOS 8 (8.0 images for pxe boot and 8 for packages).
By configuring profiles, I have been able to deploy c[001-002] with CentOS 8 and c003 with CentOS 7, from the same management server running CentOS 8.

actatux:
Hi @oxedions,

I'll try to explain the changes brought by this PR. This is a long-standing PR, I tried to answer most of the concerns raised by you and other people over the past weeks.
 
> Wouldn't that prevent users to, for example, deploy part of the cluster in RHEL 8.1
> while cluster is still in in production in RHEL 8.0, to test stability?

With the new code, it won't prevent this possibility anymore. The `$releasever` will default to the major release, but it is now possible to overwrite it with an equipment_profile parameter named `repositories_releasever` (friendly name to ease web searches). One can run a group of nodes with RHEL 8.0, another group of nodes with RHEL 8.1 and even a third group of nodes with RHEL 8 (latest).

> This will bring 2 issues for me and them:
> 1 . We loose simple minor-release granularity. So this is a regression.

This is not a regression anymore. It is possible to have minor-release granularity.

> 2 . The idea of using "facts" [...] for repositories remove ability to manually manipulate them.
> This is a very important behavior for all of us :-)

The code allows to overwrite `$releasever`, so we can manipulate it. I can patch the code to overwrite `$basearch` as well, however, as explained previously, it would not make sense.

The purpose of this PR is to allow the use of the major release in the baseurl, as CentOS and RHEL do by default. **The code allows now to use major or minor, this is win-win.**

Why am I using a new parameter `repositories_releasever` instead of `distribution_version`?

While it may seem redundant at first, it is not.
The `distribution_version` is very useful to get the path to the deployment images (_images/{vmlinuz,initrd.img,squashfs}_). It allows to deploy a node with a known technical state (eg. installer from DVD ISO 8.0).
The `repositories_releasever` is useful to force the minor release one wants to use.

To make it short:

-  `distribution_version = minor version of the boot images for PXE` 
-  `repositories_releasever = minor version of the packages to install and update`


Other topic:

>> I have seen many issues migrating from a minor to another, and some of our users 
>> sometime ask for very specific versions while the cluster is running main production
>> with another minor.
>
> I would rather manage environments to achieve this:
> http://repository_ip/repositories/centos/$environment/$releasever/$basearch/repo

This was a proposal in an open discussion, however it seems to be of interest.
With the URL above, we can define any value for `$environment`, it's as simple as '`echo production > /etc/dnf/vars/environment`'. It's easy to implement, the logic is the same than for `$releasever`.

johnnykeats:
Read all. Thank you for this PR. I will check with Ox first about all of this, there seems to be multiple devs in parallel.

johnnykeats:
@actatux. Nice PR. Will do some tests and come back to you with comments.

johnnykeats:
I think file releasever.j2 is missing. Could you add it?

actatux:
My bad! Should be ok now.

johnnykeats:
Thanks. May be wrong, but still not seeing the file targeted by
`    src: releasever.j2`

actatux:
My bad! Doing 15 things at once...I pushed to the wrong branch. It's ok (and verified) now. Sorry.

johnnykeats:
Ok now. Thanks.

resources/documentation/install_first_management.rst
```diff
@@ -52,48 +52,65 @@ Download:
...
 
 .. code-block:: bash
 
-  mkdir -p /var/www/html/repositories/centos/7.6/x86_64/os/
+  mkdir -p /var/www/html/repositories/centos/7/x86_64/os/
   mount CentOS-7-x86_64-Everything-1810.iso /mnt
-  cp -a /mnt/* /var/www/html/repositories/centos/7.6/x86_64/os/
-  umount /mnt
-  restorecon -Rv /var/www/html/repositories/centos/7.6/x86_64/os
+  cp -a /mnt/* /var/www/html/repositories/centos/7/x86_64/os/
```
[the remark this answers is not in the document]
johnnykeats: Also specify here that this major related directory will be used later to gather Centos Updates / RHEL Erratas, shared by minors images.

resources/documentation/install_first_management.rst
```diff
@@ -121,11 +144,11 @@ Repositories structure follows a specific pattern:
...
-                        |             +---+   |               |
-                        +-----------+     |   |      +--------+
-                                    |     |   |      |
-                                    v     v   v      v
-       /var/www/html/repositories/centos/7.6/x86_64/os
+                        |             +--+    |               |
+                        +-----------+    |    |    +----------+
+                                    |    |    |    |
+                                    v    v    v    v
+       /var/www/html/repositories/centos/7/x86_64/os
```
johnnykeats, line 151: Would be nice to explain here the idea of splitting image and packages.
And maybe move this explanation part before all the iso copy procedure.

roles/core/pxe_stack/files/osdeploy/redhat_8.ipxe
```diff
@@ -17,11 +17,11 @@ echo | > Target kickstart: ${eq-equipment-profile}.kickstart.cfg
 echo |
```
johnnykeats, line 17: Need to split line 15 above to echo both image and repository roots, same for centos_8, 7, etc.

roles/core/repositories_client/tasks/centos_7.yml
```diff
@@ -13,13 +13,32 @@
 
 - name: Set baseurl prefix
   set_fact:
-    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/{{ equipment_profile['operating_system']['distribution_version'] }}/{{ equipment_profile['hardware']['cpu']['architecture'] }}/"
+    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"
+
+- name: Force $releasever
```
johnnykeats, line 18: I like this.
Allows to use remote updates without updating to next minor in case of need. Then simply remove lock with next task.

Would this lock work if user do a createrepo on updates/erratas? If not, would be nice to add this warning in the documentation.

roles/core/repositories_client/templates/releasever.j2
```diff
@@ -0,0 +1 @@
+{{ equipment_profile.operating_system.repositories_releasever }}
```
johnnykeats, line 1: Don't have infra today to validate this. Did you tested it on both Centos and RHEL?

roles/core/repositories_server/readme.rst
```diff
@@ -17,17 +17,17 @@ Note that repositories structure follows a specific pattern:
 
                   Distribution    Version   Architecture    Repository
                         +             +       +               +
-                        |             +---+   |               |
-                        +-----------+     |   |      +--------+
-                                    |     |   |      |
-                                    v     v   v      v
-       /var/www/html/repositories/centos/7.6/x86_64/os
+                        |             +--+    |               |
+                        +-----------+    |    |    +----------+
```
johnnykeats, line 21: Same here, copy main doc part that explains the split between image and repos.

roles/core/pxe_stack/files/osdeploy/redhat_8.ipxe
johnnykeats: I like verbosity.

roles/core/repositories_client/tasks/redhat_7.yml
```diff
@@ -2,13 +2,32 @@
 
 - name: Set baseurl prefix
   set_fact:
-    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/{{ equipment_profile['operating_system']['distribution_version'] }}/{{ equipment_profile['hardware']['cpu']['architecture'] }}/"
+    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"
```
johnnykeats, line 5: How would this react if user overrides to uses a minor only?
Ok for one minor, since iso packages were copied in major repo, but in case of multiple minors?

johnnykeats: This is the only point worrying me.

roles/core/repositories_client/tasks/centos_7.yml
strus38: Also why do you force 'http', we should open it to https to by using a variable set by default on http if you like, but pretty sure one day https will be needed too.

roles/core/repositories_client/tasks/redhat_7.yml
strus38: Why only http and not https by adding a variable?

johnnykeats:
Please only focus on the blocking point. Once resolve, others can be seen, as minors adjustments.

roles/core/repositories_client/templates/releasever.j2
actatux: Only CentOS 7 and 8. I can test on RHEL if you want. Behaviour should be the same though.

roles/core/repositories_client/tasks/redhat_7.yml
actatux: Is this the blocking point?

actatux: If my understanding of BlueBanquise is correct, it's up to the user to import the RPMS and create the repositories on the server side. The current documentation provides the commands to create the initial minor repo (install_first_management.rst).

With the current code (master), to upgrade, a user running 7.6 may have to copy the content of the 7.7 iso to the _/var/www/html/repositories/centos/7.7/_ directory and to change `distribution_version` to 7.7 in the inventory.

> How would this react if user overrides to uses a minor only?

Almost the same as today, since the stack is currently minor only. :-)

> Ok for one minor, since iso packages were copied in major repo, but in case of multiple minors?

User will have to copy the content of the 7.7 iso, then to update `repositories_releasever` in the equipment profile. The user may also update `distribution_version` to use the 7.7 installer.

If a user wants to manage multiple minors for several equipment profiles, he will have to manage one repository per minor release. Might be self-explanatory if I apply the change proposed above (copy the whole iso to the minor directory).

actatux: To be clear, if a user decides to use minor-only, there is no need to create a major repository.

roles/core/repositories_client/tasks/centos_7.yml
actatux: While I like the idea, can this http/https feature request be postponed to another PR :-)

roles/core/repositories_client/templates/releasever.j2
johnnykeats: Then perfect. No need more tests.

roles/core/repositories_client/tasks/redhat_7.yml
johnnykeats: Probably missing folders map you have in mind. Please detail it.
In my mind:

Major versioned cluster:

/repositories/centos/8/x86_64/ <-- here goes packages trans-minors (updates)
/repositories/centos/8.1/x86_64/os <-- here os minor image

Minor versioned cluster:

/repositories/centos/8.1/x86_64/os <-- here os minor image and base repo
/repositories/centos/8.1/x86_64/ <-- here other minor related packages (updates for minor)

In second case (minor versioned)
`    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"
`
would bind to 8 even when using `repositories_releasever `.

This is the issue I see. Maybe another folder map would solve it.

johnnykeats: Solution could be to force static path on client for minor versioned system and use dynamic path on major versioned system. Would make sense.

`baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/{% if minor... %}static{% else (default) %}$releasever/$basearch/{% endif %}"`

johnnykeats:
Found it. Testing then will write proposal.

roles/core/repositories_client/tasks/redhat_7.yml
actatux: This is not a cluster-wide major/minor, this is for repositories only:

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0
    #repositories_releasever: "8.1"
```

/repositories/centos/8/x86_64/os/ <-- here go packages (updates BaseOS+AppStream)
/repositories/centos/8.0/x86_64/os/ <-- here os minor image

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0
    repositories_releasever: 8.0
```

Same directory:
/repositories/centos/8.0/x86_64/os/ <-- here go packages (updates 8.0 BaseOS+AppStream)
/repositories/centos/8.0/x86_64/os/ <-- here os minor image

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0
    repositories_releasever: 8.1
```

/repositories/centos/8.1/x86_64/os/ <-- here go packages (updates 8.1 BaseOS+AppStream)
/repositories/centos/8.0/x86_64/os/ <-- here os minor image

actatux: Are you used to CentOS 7 with `base` and `updates`? This might be the source of your question.
[Update repo disappeared completely in CentOS 8](https://wiki.centos.org/FAQ/CentOS8#I_don.27t_see_the_updates_repo_for_CentOS-8).
There is no such split in RHEL 7, nor in CentOS/RHEL 8. All the updates are cumulative.

People running CentOS 7 may indeed have to sync both `base` and `updates` to `/repositories/centos/8.1/x86_64/os/` if the iso was copied in 8.0. This is specific to CentOS 7.

actatux: Second though, I would sync `base` to `/repositories/centos/7.7/x86_64/os/` and `updates` to `/repositories/centos/7.7/x86_64/updates/`, and add `updates` to repositories for equipments running CentOS 7.
How would you achieve this with current master code?

johnnykeats: Yes, Centos 7 and Ubuntu users in our laboratory.

Still, with second and third example above

> ```
>   operating_system:
>     distribution: centos
>     distribution_major_version: 8
>     distribution_version: 8.0
>     repositories_releasever: 8.0
> ```
> 
> Same directory:
> /repositories/centos/8.0/x86_64/os/ <-- here go packages (updates 8.0 BaseOS+AppStream)
> /repositories/centos/8.0/x86_64/os/ <-- here os minor image
> 
> ```
>   operating_system:
>     distribution: centos
>     distribution_major_version: 8
>     distribution_version: 8.0
>     repositories_releasever: 8.1
> ```
> 
> /repositories/centos/8.1/x86_64/os/ <-- here go packages (updates 8.1 BaseOS+AppStream)
> /repositories/centos/8.0/x86_64/os/ <-- here os minor image

How could /$releasever/$basearch/ work on client side? This is my question. As in these two examples, if got your example correctly, there is no major repo.

Found another approach, only adjusting yours and pushing major to max. Let me detail this:

actatux: By default, dnf/yum resolve $releasever = major.
If user defines `repositories_releasever: 8.1`, the task "Force $releasever" will write the value 8.1 to /etc/dnf/vars/releasever, thus dnf/yum will resolve $releasever = 8.1.
This is documented [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sec-Configuring_Yum_and_Yum_Repositories#sec-Using_Yum_Variables).

johnnykeats: Like:
- major version as default. Sounds logical (Centos or Ubuntu, not sure for OpenSuse. @oxedions ?)
- the minor lock when using the major way, to prevent issues with drivers/km like Nvidia ones
- the split between image and packages, very didactic
- the /$releasever/$basearch for client side when using major, as it make sens and is the standard way
- the environment capability. Could be coupled with git versioning of inventory

Dont like:
- presence of minor related repo in major only version. Major way should be pushed max

Need:
- major for Centos 8, after reading your link
https://wiki.centos.org/FAQ/CentOS8#I_don.27t_see_the_updates_repo_for_CentOS-8
@oxedions also confirmed me the need in its job. Looking at this PR, seems you also need it?
- the lock on major, to prevent production unexpected update behavior
- minors, but not only as lock, as some of our software versions (ANSYS for example) only work properly with dedicated minors Centos versions, and so need often redeploy nodes on the fly depending of need. Need in my job

Proposal:

# Default is major

## Basic

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
```

No need to describe minor most of the time (could be Centos 7, 8, Ubuntu 20.04, etc). Major is pushed max.
Image and packages are fully separated:

```
/repositories/centos/8/x86_64/base/  <-- here goes copy of iso packages, and after copy of updates/erratas
/repositories/centos/8/x86_64/image/   <-- here goes image, don't care about minor used as source. system admin duty
```

On client side:

`    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"`

Simple, robust.

## With lock

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_minor_lock: 8.0
```

Same as previous, only here trigger the very nice lock added. Once drivers/kernel modules related or other elements are ready for next minor, comment the lock and replay role (or update lock to next minor).

# Minor, not default

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0
```

Added non mandatory variable `distribution_version`. Trigger minor release mechanism.
Would prefer calling this variable distribution_minor_version, but does not comply with Ansible facts.

Now, same as before, but minor related repos:

```
/repositories/centos/8.0/x86_64/base/  <-- here goes copy of iso packages, and after copy of updates/erratas for this 8.0 version (if 7.x, can add an update repo in inventory)
/repositories/centos/8.0/x86_64/image/   <-- here goes image of 8.0 iso
```

`distribution_minor_lock` is deactivated in this case.

On client, static, as desired by user:

`    baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/{{ equipment_profile['operating_system']['distribution_version'] }}/{{ equipment_profile['hardware']['cpu']['architecture'] }}/"`

# Environment

Simply detect non mandatory variable `repositories_environment` and add it on paths. Minor dev, big advantages

Would that match all requirements?
@actatux @strus38 @oxedions ?

johnnykeats: If ok, there is no reason you should handle all the development, this is a huge work (doc).
Can create a dedicated branch for this important change, PR pushed inside, and I can help.

actatux: Can you explain me why are you calling it a lock? Just want to be sure we are talking about the same thing.
$releasever is not a lock, it's a variable to resolve the URL. We can override the variable to resolve to another URL.

actatux: I can record a demo is it may help.

johnnykeats: > Can you explain me why are you calling it a lock? Just want to be sure we are talking about the same thing.
> $releasever is not a lock, it's a variable to resolve the URL. We can override the variable to resolve to another URL.

Was referring to this link: Locking a Linux System to a Specific OS Version, https://avinetworks.com/docs/18.2/how-to-tie-a-system-to-a-specific-update/

So to this code
```
  template:
    src: releasever.j2
    dest: /etc/yum/vars/releasever
    owner: root
    group: root
    mode: 0644
  when: equipment_profile.operating_system.repositories_releasever is defined
  tags:
    - template
```

Not to $releasever, used in the client dynamic URL.

In my example
`    distribution_minor_lock: 8.0
`

Would trigger the /etc/yum/vars/releasever file. Nothing related to the repositories client side path.

actatux: > And inside the file: {{ equipment_profile.operating_system.distribution_minor_lock }}
> Not to $releasever, used in the client dynamic URL.

In your example, the value of equipment_profile.operating_system.distribution_minor_lock is the value of $releasever on the client.

Repo file:
```
[root@c003 ~]# grep baseurl /etc/yum.repos.d/os.repo 
baseurl = http://10.10.0.1/repositories/centos/$releasever/$basearch/os/
```

Default, no override of $releasever:
```
[root@c003 ~]# ls /etc/yum/vars/                     
contentdir  infra
[root@c003 ~]# yum repoinfo os | grep Repo-baseurl   
Repo-baseurl : http://10.10.0.1/repositories/centos/7/x86_64/os/
```

Override $relesever:
```
[root@c003 ~]# echo 7.5 > /etc/yum/vars/releasever   
[root@c003 ~]# yum repoinfo os | grep Repo-baseurl
Repo-baseurl : http://10.10.0.1/repositories/centos/7.5/x86_64/os/
```
Rollback to default:
```
[root@c003 ~]# rm -f /etc/yum/vars/releasever
[root@c003 ~]# yum repoinfo os | grep Repo-baseurl
Repo-baseurl : http://10.10.0.1/repositories/centos/7/x86_64/os/
```

> Nothing related to the repositories client side path.

I'm sorry, I don't understand. May need a chat or a talk. The $releasever is definitely related to the repositories client side path.

actatux: I believe I got it, _Locking a Linux System to a Specific OS Version_ is confusing.
Did you believe setting a minor release to /etc/yum/vars/releasever on the client side would allow to use a major Repo-baseurl : http://10.10.0.1/repositories/centos/7/x86_64/os/ while locking to some packages version?

johnnykeats: Yes.

Thought it was equivalent to
`yum --releasever=6.6 update`

See https://access.redhat.com/solutions/238533

So it only affect the $releasever variable, by manually setting it.
Thanks clarifying that point.

Then we can drop this parameter for now. On my side, the proposal I made from your PR, i.e.:

- full major (default, only one major repo with image from any minor iso and base from updates/erratas) and "dynamic" path on client (/$releasever/$basearch)
- full minor (not default, one repo per minor) and static path from inventory

is enough.

Ok for you this way?
I will correct it now, see above.

actatux: We are on the same page, great. :)
Let me comment below.

actatux: > Dont like:
> presence of minor related repo in major only version. Major way should be pushed max

I tend to think like you, but received interesting feedback which made me change my mind.

First, `distribution_version` is currently mandatory and must exist in every inventory, we can continue this way. To install a cluster with BlueBanquise, a user should follow the documentation and copy the boot images to the minor release directory matching the DVD iso. Most of the time, there is no need to change the boot images when a new minor is available, one can still use the installer of previous minor release.

Second, the boot images are tightly coupled to the minor release since they are delivered within the iso. It makes sense to keep a copy of each boot images in a clearly identified directory. (8.0, 8.1, etc).

Third, some users need to run the installer with a known technical state: "I know nodes in equipment profile A will deploy with anaconda installer from 8.0 boot images, I don't want to change it". However, new equipments, in equipment profile B, may need newer boot images (new kernel to support new drives, CPU). They still want to install the latest packages available in the major repo, or even packages from another minor repo. (@all: I want to emphasize the difference between installer images in `images-root` and source of packages to install in `repository-root`.)

As you noticed, there is a split between images and packages. This PR currently help to use major repository, which is the default in CentOS/RHEL (at least). We can implement major release in `images-root`, but I'm afraid it will lead to confusion on the user side who will continue to believe boot images and packages are tightly coupled. Let's wait feedback from other reviewers.

## Default is major

>  Image and packages are fully separated:
> `/repositories/centos/8/x86_64/base/  <-- here goes copy of iso packages, and after copy of updates/erratas`

No problem for me but it would break existing setup and the use of "magic" repo `os` in the inventory.
Depending on the feedback on the topic above, image and packages could be fully separated with:

```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0  # used in pxe_stack only to know which image to boot
```

Currently, the 3 parameters are mandatory.

```
/repositories/centos/8/x86_64/os/  <-- here goes copy of iso packages, and after copy of updates/erratas
/repositories/centos/8.0/x86_64/images/   <-- here goes image, system admin duty to add other minors if requested and update/add environment accordingly.
```

> On client side:
>
> `baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"`
>
> Simple, robust.

I do agree. :-)

## Minor, not default

My proposal:
```
  operating_system:
    distribution: centos
    distribution_major_version: 8
    distribution_version: 8.0     # used in pxe_stack only to know which image to boot
    repositories_releasever: 8.1  # used to overwrite $releasever on the clients of this equipment profile. Could be the same than distribution_version.
```

Same as before, with minor related repo:
```
/repositories/centos/8.1/x86_64/os/  <-- here goes copy of iso packages, and after copy of updates/erratas for this 8.1 release
/repositories/centos/8.0/x86_64/images/   <-- here goes image, system admin duty to add other minors if requested and update/add environment accordingly.
```

On client side, use the same baseurl as for major repo. Purpose is to define a single baseurl_prefix in the role to keep it simple:
`baseurl_prefix: "http://{{ networks[j2_node_main_network]['services_ip']['repository_ip'] }}/repositories/{{ equipment_profile['operating_system']['distribution'] }}/$releasever/$basearch/"`

But override `$releasever` with value of `repositories_releasever` in /etc/dnf/vars/releasever, so dnf/yum will be able to use the correct URL. I chosed the name of the variable to be as close as possible to the behaviour it changes. It's also a friendly name to find related upstream documentation. You can search "repositories_releasever" on your preferred search engine.

To summarize my proposal, focused on dnf/yum repos only:

- major repo only (default, only one major repo with packages from latest minor iso and packages from updates/errata), "dynamic" path on client (/$releasever/$basearch), keep minor images for pxe deployment
- minor repo (not default, one repo per minor), "dynamic" path on client (/$releasever/$basearch), same or different images for pxe deployment

actatux: Thinking about it, your major only proposal is of interest as well for people who need simplicity. I'm wondering if we could implement the major only version (packages + images) with 2 PR?
This PR for major packages repo.
A new PR for major images.
(+ another PR to add feature to manage $environment.)

With the inventory below:
 ```
operating_system:
    distribution: centos              # mandatory
    distribution_major_version: 8     # mandatory
    distribution_version: 8.0         # optional, used in pxe_stack only to know which image to boot
    repositories_releasever: 8.1      # optional, used to override $releasever on the clients of this equipment profile. Could be the same than distribution_version.
    repositories_environment: staging # optional, used to override $environment on the clients of this equipment profile.
```

User could either:

- define mandatory parameters only: use major release for images and packages, for admins who want simplicity. The install documentation should default to this.
- define optional distribution_version: use a well-known technical state to run anaconda installer
- define optional repositories_releasever: use a repo with packages from the given minor update to ensure compatibility with soft/drivers.
- define optional repositories_environment: for admins who need to manage several environments (prod/staging/whatever)
- combine any optional parameters to achieve any combination

Comment (johnnykeats) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: Last proposal is perfect, and I understand your arguments. Ok on my side, and of for multiple PR.
Will simplify default documentation, which is the target.
Also major as default comply more with other Linux distributions (stack target is not high performance computing, but to be generic).

One last thing, non blocking:

> Third, some users need to run the installer with a known technical state: "I know nodes in equipment profile A will deploy with anaconda installer from 8.0 boot images, I don't want to change it".

Then would it make sens to allow optionally to bind anaconda OS repository to the one provided in minor (iso copy), instead of letting installer use the major repository and so load latest packages? If idea is to ensure a technical state.
Bad side: need to update and probably reboot node a second time after PXE process.
Idea only, no need to consider it now.

Makes me think stack does not provide a simple role that "update" system to latest available packages (with an exclude option available in inventory). If a new node PXE with major, this node may not be synchronized with the others if no role ensure update of all hosts.
Not sure a role is a good thing for this, more system admin task. Advice?

As you already did a lot, if you need help (documentation?) please ask.

Comment (actatux) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: > Last proposal is perfect, and I understand your arguments. Ok on my side, and ok for multiple PR.

Great. Can I answer other non blocking comments on this PR then? (will do later today, out for the afternoon)

> Then would it make sens to allow optionally to bind anaconda OS repository to the one provided in minor (iso copy), instead of letting installer use the major repository and so load latest packages?

Yes. Need to think about it.
Should we open an issue with tasks to track all these question and possible enhancements?
It would allow to review this PR and move forward.

Comment (johnnykeats) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: Please freeze this topic until @oxedions validate this. So you don't lose time.

Then yes, lets see non blocking parts, track all in a dedicated issue, and move forward.

Comment (actatux) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: > Makes me think stack does not provide a simple role that "update" system to latest available packages (with an exclude option available in inventory). If a new node PXE with major, this node may not be synchronized with the others if no role ensure update of all hosts.
Not sure a role is a good thing for this, more system admin task. Advice?

You're right, that's definitely an admin task to run dnf check-update/dnf update. It's easy to scale with clush.
Most of the time major repo won't change during production because sysadmins won't sync repositories from upstream on a daily basis, but during a maintenance window instead. Or they will perform rolling updates where a group of nodes will be removed from the job scheduler upon job completion, updated and added to a new partition.

Comment (johnnykeats) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: Thanks for this experience sharing.

**Comment by strus38** [comment]
And why don't you care about optional https capability?

Le dim. 26 avr. 2020 à 21:08, johnnykeats <notifications@github.com> a
écrit :

**Comment by johnnykeats** [comment]
> And why don't you care about optional https capability?

Do care. However https brings certificate need. This PR is already complex (to match all needs), it is better to not add more complexity. Can be done later.

**Comment by strus38** [comment]
Ok, I would have introduced the 'http_protocol' variable now, but ok, I
understand.
Thanks

Le lun. 27 avr. 2020 à 09:44, johnnykeats <notifications@github.com> a
écrit :

**Comment by oxedions** [comment]
Hello :-)

It is perfect for me this way. Stack remains simple, and we all get what we needed.

I validate this. We can proceed with minor discussions. I will comment on my side.

File: roles/core/repositories_client/tasks/redhat_7.yml (hunk 7 above)
Comment (oxedions) [reply roles/core/repositories_client/tasks/redhat_7.yml:5]: Ok for me. This is perfect :-)

File: roles/core/repositories_client/tasks/centos_7.yml (hunk 4 above)
Comment (oxedions) [reply roles/core/repositories_client/tasks/centos_7.yml:18]: I agree on this. Https will have to be handled at some point, but please later.

File: resources/documentation/install_first_management.rst (hunk 1 above)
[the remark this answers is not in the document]
Comment (oxedions) [reply resources/documentation/install_first_management.rst:61]: I like the idea of splitting both. Make things simples, and as the stack as a didactic aspect, it makes new users understand the idea behind each part.

**Comment by oxedions** [comment]
This makes me think : please add yourself in the role readme, in logs. Also upgrade role version number in vars :-)

File: resources/documentation/install_first_management.rst [hunk 8]
```diff
@@ -52,48 +52,65 @@ Download:
...
 
 **Centos/RHEL 7:**
 
+Copy the boot images to the web server directory. The path depends on the minor
+release of the DVD iso:
+
+.. code-block:: bash
+
+  mkdir -p /var/www/html/repositories/centos/7.6/x86_64/os/
+  cp -a /mnt/{.treeinfo,images,LiveOS} /var/www/html/repositories/centos/7.6/x86_64/os/
```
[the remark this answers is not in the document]
actatux: Done

resources/documentation/install_first_management.rst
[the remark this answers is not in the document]
actatux: Done

resources/documentation/install_first_management.rst
actatux: Done

roles/core/pxe_stack/files/osdeploy/redhat_8.ipxe
actatux: Done

roles/core/repositories_server/readme.rst
actatux: Done

actatux:
Code updated following your feedback.
I changed the base branch from `master` to `feat/182/enhance_repositories`. Note this branch is not protected from direct merge of PR.

actatux:
Reminder from previous resolved conversation:

This PR for major packages repo.
\+ a new PR for major images.
\+ another PR to add feature to manage $environment.
\+ another PR to move this role to advanced-core and create a simple core role.

Once all PR will be merged in `feat/182/enhance_repositories`, we can merge the branch to `master`.

With the inventory below:
 ```
operating_system:
    distribution: centos              # mandatory
    distribution_major_version: 8     # mandatory
    distribution_version: 8.0         # optional, used in pxe_stack only to know which image to boot
    repositories_releasever: 8.1      # optional, used to override $releasever on the clients of this equipment profile. Could be the same than distribution_version.
    repositories_environment: staging # optional, used to override $environment on the clients of this equipment profile.
```

User could either:

- define mandatory parameters only: use major release for images and packages, for admins who want simplicity. The install documentation should default to this.
- define optional distribution_version: use a well-known technical state to run anaconda installer
- define optional repositories_releasever: use a repo with packages from the given minor update to ensure compatibility with soft/drivers.
- define optional repositories_environment: for admins who need to manage several environments (prod/staging/whatever)
- combine any optional parameters to achieve any combination

johnnykeats:
Perfect. Thanks. Branch protected (@actatux @oxedions @johnnykeats can push only).

PR #154, opened 2020-04-08
feat: security.ssh.hostkey_checking=true|false

actatux:
Set security.ssh.hostkey_checking=false to apply the following
parameters to the hosts in /root/.ssh/config:

  StrictHostKeyChecking no
  UserKnownHostsFile=/dev/null

Default to false if not defined to keep backward compatibility with existing setup.
Set to true in example inventories for new setup.

### DIFF
--- a/roles/core/ssh_master/templates/config.j2
+++ b/roles/core/ssh_master/templates/config.j2
@@ -3,8 +3,8 @@
 {# Macro to avoid redundancy #}
 {% macro write_host(host,macro_node_main_network) %}
 Host {{host}}
-    {% if ssh_master_disable_hosts_check is defined and
-          ssh_master_disable_hosts_check == true %}
+    {% if security.ssh.hostkey_checking is not defined or
+          security.ssh.hostkey_checking == false %}
     StrictHostKeyChecking no
     UserKnownHostsFile=/dev/null
     {% endif %}
--- a/roles/core/ssh_master/templates/config.j2
+++ b/roles/core/ssh_master/templates/config.j2
@@ -1,9 +1,13 @@
+#jinja2: lstrip_blocks: "True"
 {% import "macros/network.j2" as macro_network with context %}
 {# Macro to avoid redundancy #}
 {% macro write_host(host,macro_node_main_network) %}
 Host {{host}}
+    {% if ssh_master_disable_hosts_check is defined and
+          ssh_master_disable_hosts_check == true %}
     StrictHostKeyChecking no
     UserKnownHostsFile=/dev/null
+    {% endif %}
     Hostname %h-{{macro_node_main_network}}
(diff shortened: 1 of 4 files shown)


oxedions:
And thanks for documenting it :-)

oxedions:
@bouriquet @hmeAtos Do you validate this ?

roles/core/ssh_master/templates/config.j2
```diff
@@ -1,9 +1,13 @@
+#jinja2: lstrip_blocks: "True"
 {% import "macros/network.j2" as macro_network with context %}
 {# Macro to avoid redundancy #}
 {% macro write_host(host,macro_node_main_network) %}
 Host {{host}}
+    {% if ssh_master_disable_hosts_check is defined and
+          ssh_master_disable_hosts_check == true %}
```
bouriquet, line 7: setting disable to true is quite disturbing, can you rename this variable to something more accurate like :
ssh_master_hostkey_check = true|false

Or 

enable_hostkey_check

Also, it should be present in the inventory (with hostkey check set to no to ensure backward compatibility)

actatux: I want to make it clear to the sysadmin that they are disabling hostkey check, which is a security concern. `ssh_master_disable_hostkey_check` might be better wording though.

> Also, it should be present in the inventory (with hostkey check set to no to ensure backward compatibility)

I don't want it to be backward compatible, hence the `breaking change` label. I believe it's better to provide more security by default and let user remove security on purpose.

oxedions: I understand both.

Please give me few hours, I might have an idea regarding this and the future of the security side of the stack. :-)

oxedions: A new file, group_vars/all/security.yml will be added.

```
---
security:
  ssh:
    hostkey_check: true # enable or disable hostkey check
```

Since many security features will come with next releases, having a dedicated file is better for sys admin to understand the risks.

actatux: So no `disable` in the variable name to explicit the disabling of a security mechanism?

oxedions: I think the fact that these parameters are set in this file, called security, will warn users manipulating these parameters has to be done with care.
We should even add an ASCII Warning art, to inform this file is security key.

oxedions:
Many thanks to you guys :-)

roles/core/ssh_master/templates/config.j2
actatux: nagios values are return codes, not boolean. Any application in Unix should return 0 when it completes successfully.
Anyway I updated the PR to use security.ssh.hostkey_checking.

PR #155, opened 2020-04-08
feat: Use Python's logging with bootset

actatux:
The output may be too verbose when running the bootset command for
hundreds or thousands of nodes. Let the sysadmin choose to quiet the
output.

Switch from print() to logging and add a new parameter --quiet to set
the logging level to WARNING.

### DIFF
--- a/roles/core/pxe_stack/files/bootset.py
+++ b/roles/core/pxe_stack/files/bootset.py
@@ -14,10 +14,11 @@
 
 from ClusterShell.NodeSet import NodeSet
 from argparse import ArgumentParser
-import yaml
+import grp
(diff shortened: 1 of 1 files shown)


oxedions:
I like the idea.
Do you confirm that using:

`     loglevel = logging.WARNING`

Then all logging.error will still be displayed ?

actatux:
Yes. You can refer to https://docs.python.org/3/howto/logging.html

> The default level is WARNING, which means that only events of this level and above will be tracked, unless the logging package is configured to do otherwise.

I set the default logging level to INFO to keep the current behaviour.

Issue #146, opened 2020-03-26
dns_server role: remove "@ IN A 10.0.0.1  mngt IN A 10.0.0.1" from forward template


jpm38:
These 3 lines are not needed and are leading the the dns_server role failing in chroot context execution.

diff -r roles/core/dns_server/templates/forward.j2 roles.bkp/core/dns_server/templates/forward.j212a13,15> @ IN A {{network_interfaces[j2_node_main_network_interface]['ip4']}}>> {{inventory_hostname}} IN A {{network_interfaces[j2_node_main_network_interface]['ip4']}}
--

actatux:
Hello @jpm38 , there is a proposal in #158, does it solve your issue?
PR is still pending for reviews.

jpm38:
Hello @actatux 
Yes, it's perfect.
thanks a lot

PR #158, opened 2020-04-09
fix: improve DNS forward and reverse templates

actatux:
Fixes several caveats with the zone templates:

 - Add $ORIGIN
 - Remove useless Address record for the domain
 - Remove inventory_hostname defined twice (explicitly + in the loop)
 - Remove external_hosts, cannot be defined in the zone

Closes: #146

### DIFF
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/external.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/external.yml
@@ -14,7 +14,7 @@ external_dns:
   dns_client: # set directly on client side in resolv.conf
     - 208.67.220.220
 
-# Hosts defined here will be automatically added into /etc/hosts and into DNS configuration
+# Hosts defined here will be automatically added into /etc/hosts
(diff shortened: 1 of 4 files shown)


oxedions:
Hello @actatux 

So external hosts will not be added into the DNS.
Is it expected ?

actatux:
It can't work.

```
[root@management1 bluebanquise]# nslookup sphenisc.com 10.10.0.1
Server:         10.10.0.1
Address:        10.10.0.1#53

** server can't find sphenisc.com: REFUSED

[root@management1 bluebanquise]# nslookup sphenisc.com.tumulus.local. 10.10.0.1
Server:         10.10.0.1
Address:        10.10.0.1#53Name:   sphenisc.com.tumulus.local
Address: 213.186.33.4
```

actatux:
Adding them in /etc/hosts should do the trick anyway.

oxedions:
Then it is ok for me.

oxedions:
@bouriquet @hmeAtos ? Are you ok with this fix ?

PR #160, opened 2020-04-10
docs: configure SSH public key before OS deployment

actatux:


resources/documentation/deploy_bluebanquise.rst
```diff
@@ -148,6 +148,18 @@ Note also that if you update configuration on management1, it is recommanded to
...
+--------------
+
+In order to log into the remote nodes without giving the password, check that
+the ssh public key defined in authentication.yml in your inventory match your
+management1 public key. If not, update it. Remember to run the pxe_stack role
+after updating the configuration.
+
+.. code-block:: bash
+
+  bluebanquise-playbook -p management1 -n management1 --tags pxe_stack
```
oxedions, line 161: I would prefer to keep ansible playbook command in main documentation, as bluebanquise-playbook is an additional tool, for production usage designed to facilitate usage.

actatux: > every one used to Ansible can understand the whole stack quickly :-)

The project has some specific that don't follow Ansible guidelines, it's not that easy to understand. :-)

Anyway, your comment makes sense, I will update the PR.

oxedions:
@hmeAtos @bouriquet ? Ok for you ?

resources/documentation/deploy_bluebanquise.rst
```diff
@@ -148,6 +148,18 @@ Note also that if you update configuration on management1, it is recommanded to
 
 Please refer to the pxe_stack role dedicated section in this documentation for more information on the bootset usage.
 
+SSH public key
+--------------
+
+In order to log into the remote nodes without giving the password, check that
+the ssh public key defined in authentication.yml in your inventory match your
+management1 public key. If not, update it. Remember to run the pxe_stack role
+after updating the configuration.
```
hmeAtos, line 157: I would have added the role ssh_slave, for those he uses it. No ?

actatux: At this step of the documentation, we are deploying our first nodes in the cluster. We need to update the kickstart configuration (templates/kickstart.cfg.j2) to insert the new SSH public key, which will allow to authenticate to the deployed hosts without password authentication.
Running ssh_slave may add the public key into the authorized_keys file of management1. Since we already ran Ansible successfully, it should already be configured. Am I wrong?

hmeAtos: indeed, at this step it is useless to indicate this way.
Good for me.

PR #161, opened 2020-04-10
Backport repositories configuration

bouriquet:
(cherry picked from commit ec9e900ce740390054df49636262183a6a8ab68f)


oxedions:
This PR is not valid, but as @bouriquet Asked for a special act on this no more supported branch, and as his team is the only one known to still use this version, I will accept it.
However, please note that this is a very special case, and it is not the standard.

PR #162, opened 2020-04-10
ci: introduce unit testing of roles with Molecule

strus38:
- Start with roles `time` and `dns_server`
- Added molecule lint/converge
- Remove unnecessary empty lines to pass the linter phase


actatux:
By the way, the cause of the conflicts is because your commit is not on top of master:
```
$ git log --oneline --no-merges --name-only strus38/feature/molecule..master 
629db3f (HEAD -> master, origin/master, origin/HEAD) [advanced_dhcp_sever] remove references to ntp_ip variable (#153)
resources/documentation/configure_bluebanquise.rst
resources/examples/multi_icebergs_cluster/inventory/group_vars/all/networks/ice1-1.yml
resources/examples/multi_icebergs_cluster/inventory/group_vars/all/networks/ice2-1.yml
resources/examples/multi_icebergs_cluster/inventory/group_vars/all/networks/ice3-1.yml
resources/examples/multi_icebergs_cluster/inventory/group_vars/all/networks/ice3-2.yml
roles/advanced-core/advanced_dhcp_server/templates/dhcpd.networks.conf.j2
3a34479 fix: install jinja2 macros in rpm
bluebanquise.spec
fed7da3 fix: require clustershell in specfile
bluebanquise.spec
f7ffaf0 Fix: Adding missing role Ansible
roles/core/ansible/readme.rst
roles/core/ansible/tasks/main.yml

```

strus38:
I had to do a dos2unix on all files to have the linter working.

Clear for the master branch, normally I never work on the master branch,
since for me it should always be reflecting the last stable version.

I will change my name.

Thanks

strus38:
Hi

I fixed some of your requests:
- change root to my id
- re-sync with master to avoid conflicts
- the dos2unix had to be passed on all for yamlint/ansible-lint to pass on the yaml files (so I did not cganged that)
- the test_hosts_file(host) is there by default when you init molecule within a role: ` molecule init scenario -d docker --verifier-name testinfra`, but as you can see, this part is not called yet since the 'verifier' phase is currently disabled in molecule since none of the role are currently running correctly in the converge phase because of lack of default variables. So, I want to keep that to help next people an exemple of what they have to do in such file.

If that is all, I guess you can attempt another review. Thanks.

strus38:
Note: on windows10 WSL running Ubuntu 18.04, you must define in your Ubuntu bash:
`git config --global core.autocrlf input ` to prevent LF/CRLF issues.

So I am reworking the PR to just validate the approach on 2 roles.

roles/core/time/handlers/main.yml
```diff
@@ -7,4 +7,4 @@
   loop: "{{ services_to_start[(ansible_distribution|lower|replace(' ','_'))]['_'+ansible_distribution_major_version] }}"
   when:
     - "'service' not in ansible_skip_tags"
-    - (start_services | bool)
+    - (start_services | default('yes') | bool)
```
actatux, line 10: I would define the default value in defaults/main.yml.

roles/core/time/molecule/default/converge.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Converge
+  hosts: all
+
+  tasks:
+    - name: "Include time for servers"
+      include_role:
+        name: "time"
+      vars:
+        - external_time:
```
actatux, line 10: I'm wondering if we should add these values to defaults/main.yml as well?

roles/core/time/molecule/default/molecule.yml
```diff
@@ -0,0 +1,35 @@
...
+  set -e
+  yamllint .
+  ansible-lint
+  flake8
+platforms:
+  - name: instance
+    image: "geerlingguy/docker-${MOLECULE_DISTRO:-centos8}-ansible:latest"
+    command: "/lib/systemd/systemd"
+    capabilities:
+      - "SYS_ADMIN"
```
actatux, line 16: I don't see this capability in a fresh init with molecule. Why do we need it?

roles/core/time/molecule/default/molecule.yml
```diff
@@ -0,0 +1,35 @@
...
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
```
actatux, line 18: I don't see these volumes in a fresh init with molecule. Why do we need them?

roles/core/time/molecule/default/molecule.yml
```diff
@@ -0,0 +1,35 @@
...
+    privileged: true
```
actatux, line 19: I don't see this privileged in a fresh init with molecule. Why do we need it?

roles/core/time/tasks/main.yml
```diff
@@ -21,14 +21,15 @@
...
+  when: time_zone is defined
 
 - meta: flush_handlers
 
 - name: Manage services state
   service:
     name: "{{ item }}"
-    enabled: "{{ (enable_services | bool) | ternary('yes','no') }}"
-    state: "{{ (start_services | bool) | ternary('started', omit) }}"
+    enabled: "{{ (enable_services | default('yes') | bool ) | ternary('yes','no') }}"
```
actatux, line 31: I would define the default value in defaults/main.yml.

roles/core/time/tasks/main.yml
```diff
@@ -21,14 +21,15 @@
...
+    state: "{{ (start_services | default('started') | bool) | ternary('started', omit) }}"
```
actatux, line 32: I would define the default value in defaults/main.yml.

roles/core/time/templates/chrony.conf.j2
```diff
@@ -43,12 +47,14 @@ rtcsync
 
 {% if time_profile == 'server' %}
 
-# Allow NTP client access from local network.
+# Allow NTP client access from local network.*
+{% if network_interfaces is defined %}
```
actatux, line 51: I believe `network_interfaces` is a mandatory parameter. Can you confirm @oxedions ?

roles/core/time/tasks/main.yml
```diff
@@ -21,14 +21,15 @@
 - name: Set timezone
   timezone:
     name: "{{ time_zone }}"
+  when: time_zone is defined
```
actatux, line 24: If we define time_zone in defaults/main.yml, then we would not need to add this test.

roles/core/time/templates/chrony.conf.j2
```diff
@@ -3,26 +3,30 @@
 ## {{ansible_managed}}
 
 # External servers/pool if asked
-{% if external_time.time_server.server and time_profile == 'server' %}
+{% if external_time is defined and external_time.time_server is defined and time_profile == 'server' %}
```
actatux, line 6: If we define external_time in defaults/main.yml, then we would not need to add this test.

roles/core/time/molecule/default/molecule.yml
strus38: To limit the scope of the extended privileges, grant SYS_ADMIN capabilities along with the same image...
So basically since many of the role are 'admin' like, and when we play with systemd and others, this is the recommended way. I can try without if needed, but that doesn't hurt anyway :-)

roles/core/time/molecule/default/converge.yml
strus38: We can, here it would just overwrite the defaults to test another case of needed.

roles/core/time/molecule/default/molecule.yml
actatux: Great, let's keep this. Thanks for the explanation.

roles/core/time/.ansible-lint
```diff
@@ -0,0 +1,3 @@
+skip_list:
+  - '306'
```
actatux, line 2: same as for dns_server above.

roles/core/time/templates/chrony.conf.j2
```diff
@@ -43,7 +39,7 @@ rtcsync
 
 {% if time_profile == 'server' %}
 
-# Allow NTP client access from local network.
+# Allow NTP client access from local network.*
```
actatux, line 42: typo?

roles/core/time/molecule/default/verify.yml
```diff
@@ -0,0 +1,19 @@
+---
+# This is an example playbook to execute Ansible tests.
+
+- name: Verify
+  hosts: all
+  vars:
+    contents: "{{ lookup('file', '/etc/timezone') }}"
```
actatux, line 7: The check can fail depending on the OS: there is no `/etc/timezone` in CentOS 8. The [timezone module](https://docs.ansible.com/ansible/latest/modules/timezone_module.html) uses `timedatectl` to set the time zone.

strus38:
Fixed all the remarks. Should be ok now. Thanks

strus38:
Travis result:
image
The failing ones are because 'bind' package is not available. Please, suggest how to fix it.

roles/core/time/templates/chrony.conf.j2
```diff
@@ -3,22 +3,18 @@
 ## {{ansible_managed}}
 
 # External servers/pool if asked
-{% if external_time.time_server.server and time_profile == 'server' %}
+{% if time_profile == 'server' %}
```
actatux, line 6: chrony configuration is either server or pool. Using both is redundant. (https://chrony.tuxfamily.org/faq.html#_what_is_the_minimum_recommended_configuration_for_an_ntp_client).
It looks like a bug because we defined both in the examples inventories. Can you revert the change on this file to keep current behaviour? I will open an issue ticket.

roles/core/time/templates/chrony.conf.j2
```diff
@@ -3,22 +3,18 @@
...
 server {{external_server}}
 {% endfor %}
-{% endif %}
-{% if external_time.time_server.pool and time_profile == 'server' %}
 {% for external_pool in external_time.time_server.pool %}
 pool {{external_pool}} iburst maxsources 3
 {% endfor %}
 {% endif %}
-{% if external_time.time_client.server and time_profile == 'client' %}
+{% if time_profile == 'client' %}
```
actatux, line 14: There is no default value for `external_time.time_client.server`. Anyway, see comment above and revert these changes please.

roles/core/time/templates/chrony.conf.j2
actatux: #166

strus38:
Note: I have just completed a new github projet: [docker-ubi7-ansible](https://hub.docker.com/repository/docker/strusfr/docker-ubi7-ansible) so we will be able to update travis config with ubi7 images.
The image is available on docker hub `docker pull strusfr/docker-ubi7-ansible:latest`

oxedions:
Hello,

Thank you for this PR.
Please do not merge it, I wish to take time to investigate all of this.

Have a nice weekend

Ox

roles/core/dns_server/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# implement more default values which will come from inventory
+j2_current_iceberg_network: ice-0
```
oxedions, line 3: j2 variables should never be redefined, as they are hearth of the stack. Only manual precedence is expected.
Please note also that networks and icebergs start numbering at 1, not 0 in BlueBanquise :-)

roles/core/dns_server/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
...
+external_dns: []
+networks: []
+domain_name: .localhost
+
+start_services: true
+enable_services: true
```
oxedions, line 9: I do not approve default usage, please see https://github.com/bluebanquise/bluebanquise/pull/163

roles/core/time/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
...
+      - pool.ntp.org
+  time_client:
+    server: []
+    pool: []
+time_zone: Europe/Paris
+
+# Default server configuration
+time_profile: 'server'
+icebergs_system: false
+network_interfaces: []
```
oxedions, line 24: Same than for DNS, defaults are not allowed for this kind of variables.

roles/core/dns_server/defaults/main.yml
strus38: ok, then I will move them in the converge.yml as pre_tasks, ok?

roles/core/dns_server/defaults/main.yml
strus38: ok, then I will move them in the converge.yml as pre_tasks, ok?

roles/core/time/defaults/main.yml
strus38: ok, then I will move them in the converge.yml as pre_tasks, ok?

roles/core/dns_server/defaults/main.yml
oxedions: Yes. May I ask why you need to force j2_current_iceberg_network: ice-0 ?
Is it because you are using chroot environment ?

roles/core/dns_server/defaults/main.yml
oxedions: Yes, as long as these variables are not "seen" by the ansible-playbook or the ansible-inventory commands, it is ok :-)

roles/core/time/defaults/main.yml
oxedions: yes

strus38:
Fixed the PR as suggested by comments. Please review again.

.travis.yml
```diff
@@ -8,7 +8,8 @@ env:
   matrix:
     - MOLECULE_DISTRO: centos8
     - MOLECULE_DISTRO: centos7
-    - MOLECULE_DISTRO: ubi8
+    # Need a RHEL subscription - MOLECULE_DISTRO: ubi8
+    # Need a RHEL subscription - MOLECULE_DISTRO: ubi7
     - MOLECULE_DISTRO: ubuntu1804
```
actatux, line 13: Since we don't fully support Ubuntu yet, I would comment it as well for now.

roles/core/time/molecule/default/prepare.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Prepare
+  hosts: all
+
+  tasks:
+    - name: "update cache apt"
+      apt: update_cache=yes
+      when: ansible_os_family == "Debian"
+
+    - name: "update cache yum"
```
actatux, line 10: Not needed. We need to be as close as possible than production where we don't force a cache update.
We have to update the cache in the Ubuntu image because the maintainer of the image is running `rm -Rf /var/lib/apt/lists/*` in the Dockerfile.

roles/core/dns_server/defaults/main.yml
actatux: We don't load the example inventories with Molecule.
We either have to explicitly define mandatory variables in an inventory internal to the Molecule scenario (converge.yml), or rely on defaults values. Since usage of default values is not an option here, we have to define the mandatory variables in converge.yml.
We can define different scenarios with different values to do multiple tests. (eg. one scenario to test a role behaviour with single iceberg and another scenario to test with multi iceberg)

roles/core/time/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+---
```
actatux, line 1: My bad, the stack does not use roles defaults: https://github.com/bluebanquise/bluebanquise/pull/163#issuecomment-613333223
Please remove this file. Sorry for the extra work. :-(

strus38:
all fixed I guess.

PR #163, opened 2020-04-11
fix: define default parameters in role repositories_client

actatux:
Set a default value in the role `repositories_client` for parameters `repositories` and `disable_distro_repositories`.

Rename parameter `remove_distro_repositories` to `disable_distro_repositories` because the purpose of this parameter is to disable the repositories, not to remove them.

Add a (dirty) deprecation warning (fail+ignore_errors) to inform users.

### DIFF
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/repositories.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/all/general_settings/repositories.yml
@@ -2,4 +2,4 @@ repositories:
   - bluebanquise
   - os
 
-remove_distro_repositories: True
+disable_distro_repositories: True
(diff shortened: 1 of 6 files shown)


hmeAtos:
Good for me

oxedions:
I need to provide more details about the stack variables.

This is my fourth stack. In previous stacks, failures occurred after some time, and it was always related to increasing complexity, and non centralization of data.

In BlueBanquise, I do not want to do the same mistakes.

All data should be in the inventory.
Data in roles should only be related to the role internal mechanisms (packages names, services names, versions, etc etc).
All variables subject to precedence in some way (i.e. redefined at group_vars, in host_vars, or at ansible-playbook invocation) should be stored in group_vars/all.

This for multiple reasons, but mains are that data should be defined in a centralized place, i.e. inventory or a DB if connected to a DB. This, to avoid multi-definitions, and simplify management and debug.

I know this is not the Ansible standard, but the stack do not follow all Ansible standards (we are using merge hash_behavior, roles are not generic for external usage, etc.).

I should have provide this architecture and philosophy information much before now, my bad :-)

actatux:
OK. I will open a new PR to rename parameter `remove_distro_repositories` to `disable_distro_repositories`.

PR #164, opened 2020-04-12
ci: run Molecule test with GitHub actions

actatux:
- distro:
   - centos7
   - centos8
- role:
   - core/display_tuning
   - core/dns_client
   - core/dns_server
   - core/time
   - addons/clustershell
   - addons/users_basic

### DIFF
--- a/.github/workflows/molecule-actions.yml
+++ b/.github/workflows/molecule-actions.yml
@@ -1,0 +1,44 @@
+---
+name: Unit testing - Roles
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'roles/**'
+      - '.github/workflows/molecule-actions.yml'
+  pull_request:
+    paths:
+      - 'roles/**'
+      - '.github/workflows/molecule-actions.yml'
+
+jobs:
+  molecule_test:
+    name: ${{ matrix.role }} - ${{ matrix.distro }} - molecule test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        distro:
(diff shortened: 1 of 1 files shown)


oxedions:
I do not know Molecule, so I trust you on this.

Idea is to run deeper tests on the role, or even test deployment of them ?
If deployment, can we dot that with our free Github account ?

strus38:
Should we add also: [slack-notifier](https://github.com/marketplace/actions/slack-notify)?

actatux:
> Idea is to run deeper tests on the role, or even test deployment of them ?

This change runs unit tests on the roles, when Molecule is implemented. Each role will be tested independently from the others.

> If deployment, can we dot that with our free Github account ?

No deployment yet. I believe this is not possible with GitHub free.

> Should we add also: slack-notifier?

To notify results of GH actions? Might be a good idea, maybe in a dedicated channel?

oxedions:
Is there a way to configure molecule for the whole stack in one place ? To avoid many files in each roles, since roles are not designed to be used externally of the stack.

actatux:
No. Unit test of roles need to be unitary. It helps to detect issues like #169.
We need to do integration tests for the stack, but this is a different topic.

Note that with current implementation, any PR to a single role will trigger a molecule test for all roles.

oxedions:
Ok thanks.

We will need to ensure these files are removed from releases rpm :-)

actatux:
Indeed, nice catch! #176 for tracking.

actatux:
See the "**All checks have passed**" below. :-)

PR #170, opened 2020-04-12
ci: molecule unit test for conman

actatux:


roles/core/conman/molecule/default/molecule.yml
```diff
@@ -5,9 +5,29 @@ driver:
...
+  inventory:
+    host_vars:
+      instance:
+        j2_current_iceberg: iceberg1
+        equipment_profile:
+          equipment_type: server
+          authentication:
+            equipment_user: 'ADMIN'
+            equipment_password: 'ADMIN'
+          operating_system:
```
actatux, line 23: Should not be necessary, we can get similar information with `ansible_distribution*` facts.

oxedions: Agreed, this role can use facts.

PR #173, opened 2020-04-13
fix: use with_first_found to gather OS vars

actatux:
Search vars in:
 - `<distribution>_<major>.yml`    # eg. CentOS_8.yml
 - `<os_family>_<major>.yml`      # eg. RedHat_8.yml
 - `<distribution>.yml`            # eg. CentOS.yml
 - `<os_family>.yml`               # eg. RedHat.yml

Port code to [Ansible fact namespacing](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.5.html#ansible-fact-namespacing): ansible_* -> ansible_facts.*


roles/core/dns_server/tasks/main.yml
```diff
@@ -1,8 +1,18 @@
 ---
+- name: Gathering OS specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - files:
+        - "vars/{{ ansible_facts.distribution | replace(' ','_') }}_{{ ansible_facts.distribution_major_version }}.yml"
+        - "vars/{{ ansible_facts.distribution | replace(' ','_') }}.yml"
+        - "vars/{{ ansible_facts.os_family }}.yml"
```
oxedions, line 8: Can you sync on this with PR #179 ?

        - "vars/{{ ansible_facts.distribution | replace(' ','_') }}_{{ ansible_facts.distribution_major_version }}.yml"
        - "vars/{{ ansible_facts.os_family }}_{{ ansible_facts.distribution_major_version }}.yml"
        - "vars/{{ ansible_facts.distribution | replace(' ','_') }}.yml"
        - "vars/{{ ansible_facts.os_family }}.yml"

So all are working the same way. Then perfect.

actatux: Done

actatux:
Had to rebase this PR to solve conflict with master.

roles/core/dns_server/tasks/main.yml
```diff
@@ -1,8 +1,29 @@
...
+      skip: True
+
 - name: Install packages
   package:
-    name: "{{ packages_to_install[(ansible_distribution|lower|replace(' ','_'))]['_'+ansible_distribution_major_version] }}"
+    name: "{{ item }}"
     state: present
+  loop: "{{ packages_to_install }}"
```
bouriquet, line 26: looping over package isn't really recommended, can you supply the package list in "name" without looping ?

From Ansible loops documentation :
You can pass a list directly to a parameter for some plugins. Most of the packaging modules, like yum – Manages packages with the yum package manager and apt – Manages apt-packages, have this capability. When available, passing the list to a parameter is better than looping over the task.

oxedions:
By the way @actatux 
I would like to not update more roles with this enhancement.
I would like to see how users react to this now :-)

roles/core/dns_server/tasks/main.yml
actatux: Sure. Thanks.

actatux: Done. I did not know we could pass a list to package. Thanks for the info.

PR #174, opened 2020-04-14
Implement molecule for users basic and clustershell

strus38:
Added molecule tests for user_basics and clustershell.


roles/addons/users_basic/molecule/default/converge.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Converge
+  hosts: all
+
+  vars:
```
actatux, line 5: Open discussion: is there a recommendation to define variables to test the role in converge.yml?

I read the [Molecule documentation](https://molecule.readthedocs.io/en/latest/configuration.html#id12 ) and noticed we can define an inventory for the scenario in the Provisioner:

```yaml
provisioner:
  name: ansible
  inventory:
    group_vars:
      foo1:
        foo: bar
      foo2:
        foo: bar
        baz:
          qux: zzyzx
    host_vars:
      foo1-01:
        foo: bar
```
This is the implementation I followed for #171 ([molecule.yml](https://github.com/bluebanquise/bluebanquise/blob/49b37ea954e254c2220bf9107b87d8d677e22bab/roles/core/repositories_client/molecule/default/molecule.yml)).

We can discuss this in Slack.

roles/addons/users_basic/molecule/default/prepare.yml
```diff
@@ -0,0 +1,13 @@
+---
+- hosts: all
```
actatux, line 2: Can you use the [same layout than #175](https://github.com/bluebanquise/bluebanquise/blob/af14ffa88d7e84630c5c9ca5b0ae2d87063db29c/roles/core/display_tuning/molecule/default/prepare.yml)? (files are doing the same task, should be identical)

roles/addons/users_basic/molecule/default/prepare.yml
```diff
@@ -0,0 +1,13 @@
...
+  name: prepare the environment
+
+  tasks:
+  - name: "update cache apt"
+    apt: update_cache=yes
+    when: ansible_os_family == "Debian"
```
actatux, line 8: Please use `ansible_facts.os_family` instead (https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.5.html#ansible-fact-namespacing)

roles/addons/users_basic/molecule/default/prepare.yml
```diff
@@ -0,0 +1,13 @@
...
+
+  - name: "install openssh"
```
actatux, line 10: Why?

roles/addons/users_basic/.ansible-lint
```diff
@@ -0,0 +1,3 @@
+skip_list:
+  - '204'
+  - '403'
```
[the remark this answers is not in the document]
strus38: solved. Switched to 'present', reason being the openssh package is not installed by default in the container.

roles/addons/users_basic/molecule/default/converge.yml
strus38: as discussed in slack, I would like to postpone this conversation until we get a proper default inventory for localhost molecule tests.

roles/addons/users_basic/molecule/default/prepare.yml
strus38: done.

PR #177, opened 2020-04-14
fix: rename param used to disable the distro repos

actatux:
Rename parameter `remove_distro_repositories` to
`disable_distro_repositories` because the purpose of this parameter is
to disable the repositories, not to remove them.

Add a (dirty) deprecation warning (fail+ignore_errors) to inform users.


actatux:
Closed in favour of #185.

PR #178, opened 2020-04-14
Fixing spec file to remove CICD files from RPM

strus38:
After many tries, simply removing from the source all the files we don't want to ship is the more efficient and simple way.
rpmbuild tested and working when molecule, .ansible-lint and .yamllint are present.

### DIFF
--- a/bluebanquise.spec
+++ b/bluebanquise.spec
@@ -119,5 +119,5 @@ end}
 
 %changelog
 * Sun Feb  2 2020 Bruno Travouillon <devel@travouillon.fr>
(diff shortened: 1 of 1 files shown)


bluebanquise.spec
```diff
@@ -115,4 +119,5 @@ end}
 
 %changelog
 * Sun Feb  2 2020 Bruno Travouillon <devel@travouillon.fr>
+* Fri Apr 14 2020 strus38 <indigoping4cgmi@gmail.com>
```
actatux, line 122: Almost sure today is not Friday. Can't wait for the week-end? :D

strus38: fixed

PR #187, opened 2020-04-19
Adding molecule tests for log_client and log_server

strus38:
Molecule tests for:
- log_client
- log_server

### DIFF
--- a/.github/workflows/molecule-actions.yml
+++ b/.github/workflows/molecule-actions.yml
@@ -29,6 +29,8 @@ jobs:
           - core/time
           - addons/clustershell
           - addons/users_basic
+          - core/log_client
+          - core/log_server
     env:
       MOLECULE_DISTRO: ${{ matrix.distro }}
     steps:
--- a/.github/workflows/molecule-actions.yml
+++ b/.github/workflows/molecule-actions.yml
@@ -26,11 +26,12 @@ jobs:
           - core/display_tuning
           - core/dns_client
           - core/dns_server
+          - core/log_client
+          - core/log_server
           - core/time
           - addons/clustershell
           - addons/users_basic
-          - core/log_client
-          - core/log_server
+
     env:
       MOLECULE_DISTRO: ${{ matrix.distro }}
     steps:
--- a/roles/core/log_client/molecule/default/verify.yml
+++ b/roles/core/log_client/molecule/default/verify.yml
@@ -1,16 +1,25 @@
 ---
 - name: Verify
   hosts: all
+
+  vars:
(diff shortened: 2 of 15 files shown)


roles/core/log_client/molecule/default/verify.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Verify
+  hosts: all
+  tasks:
+  - name: Check that files exists
+    stat:
+      path: "{{ item }}"
+    register: stat_results
+    with_items:
+      - /etc/rsyslog.conf
```
actatux, line 10: `dest` is a variable in the task (even if `rsyslog_conf_path` is **always** /etc/rsyslog.conf in vars). If the value should change, this test will fail.
Moreover, there is no conditional to run this task, thus we don't need to verify the file exists if the converge did not fail. (the `template` module would fail if it could not write to `dest`)

roles/core/log_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Verify
+  hosts: all
+  tasks:
+  - name: Check that files exists
+    stat:
+      path: "{{ item }}"
+    register: stat_results
+    with_items:
+      - /etc/rsyslog.conf
```
actatux, line 10: Same comment as above for log_client.

.github/workflows/molecule-actions.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           - core/time
           - addons/clustershell
           - addons/users_basic
```
actatux, line 31: Can you move this a few lines above with core roles? I tried to order alphabetically per role category priority (core/advanced-core/addons):

- core/dns_server
- core/log_client
- core/log_server
- core/time

roles/core/log_client/molecule/default/verify.yml
strus38: Removed and made the test smarter.

roles/core/log_server/molecule/default/verify.yml
strus38: same comment. Removed and smarter tests.

strus38:
Note that github action is failing, where it works perfectly locally.
To test locally, I am running `molecule test` from the role folder (aka: roles/core/log_server for example). Not sure why the paths are skrewed by github action.

roles/core/log_client/molecule/default/verify.yml
strus38: fixed

roles/core/log_server/molecule/default/verify.yml
strus38: fixed

PR #189, opened 2020-04-21
ci: molecule unit test for nic

strus38:
- adding molecule for roles/core/nic
- fixing all molecule.yml file to let the default setup for tasks


roles/core/nic/defaults/main.yml
```diff
@@ -1 +1,2 @@
+---
 nic_nm_controlled: yes
```
actatux, line 2: @oxedions I introduced this file with ea3c977b while I did not know yet you don't want to use defaults in the roles. Should we rework this?

roles/core/nic/molecule/default/verify.yml
```diff
@@ -0,0 +1,33 @@
...
+    register: ifcfg_list
+
+  - name: assert we have 3 eth files
+    assert:
+      that: ifcfg_list.matched == 3
+
+  - name: assert eth0
+    lineinfile:
+      path: /etc/sysconfig/network-scripts/ifcfg-eth0
+      line: "IPADDR={{ hostvars[inventory_hostname]['ansible_eth0'].ipv4.address }}"
```
actatux, line 19: Can you double check the test does what you want? Does the verify fail when there is a change?

```
[root@management1 ~]# cat /etc/ansible//playbooks/test.yml
- name: Verify
  hosts: localhost

  tasks:
    - name: assert
      lineinfile:
        path: /tmp/ifcfg-eth0
        line: "IPADDR=1.2.3.4"
    - name: assert
      lineinfile:
        path: /tmp/ifcfg-eth1
        line: "IPADDR=4.3.2.1"
```
Create first file with expected IP address:
`[root@management1 ~]# echo IPADDR=1.2.3.4 > /tmp/ifcfg-eth0`

Create second file with wrong IP address:
`[root@management1 ~]# echo IPADDR=4.3.1.2 > /tmp/ifcfg-eth1`

Run the playbook:
```
[root@management1 ~]# ansible-playbook /etc/ansible//playbooks/test.yml

PLAY [Verify] **************************

TASK [Gathering Facts] *****************
ok: [localhost]

TASK [assert] **************************
ok: [localhost]

TASK [assert] **************************
changed: [localhost]

PLAY RECAP *****************************
localhost : ok=3 changed=1 unreachable=0 failed=0 skipped=0 rescued=0 ignored=0
```

Check the files:
```
[root@management1 ~]# cat /tmp/ifcfg-eth0
IPADDR=1.2.3.4
[root@management1 ~]# cat /tmp/ifcfg-eth1
IPADDR=4.3.1.2
IPADDR=4.3.2.1
```

roles/core/nic/tasks/main.yml
```diff
@@ -23,3 +23,4 @@
     mode: 0644
   with_items: "{{ network_interfaces }}"
   when: (network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none)
+  changed_when: false
```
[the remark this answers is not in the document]
strus38: Without this, the idempotence test is failing.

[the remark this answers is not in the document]
actatux: Might be a bug in the role?

roles/core/nic/molecule/default/verify.yml
strus38: Given the doc, it should, but I will double check, since now I have a doubt!

roles/core/nic/tasks/main.yml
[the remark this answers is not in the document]
strus38: I don't think so, this kind of tasks need those to be idempotent... At least what I found on Google.

roles/core/nic/molecule/default/verify.yml
actatux: OK. Depending on the result of your test, can you add a comment before the lineinfile to specify the test should fail if the task report a change?

strus38:
It is the way to make it idempotent... Otherwise, it is not

actatux:
It does not make it idempotent, it makes Ansible lying, claiming there was no change while there is one. It makes molecule happy (=green), but it does not make the role idempotent.

oxedions:
Dear @strus38 
This PR is old. Do you still wish to maintain it ?
Please take into account that the new role nic_nmcli should become now the standard NIC configuration role, and nic role should become less maintained / deprecated in the future.

strus38:
Remove it then.
Thanks

oxedions:
Ok, closing.

Thank you for the time you took on this PR anyway 🙂

PR #194, opened 2020-04-24
Fix network generation with bonding

bouriquet:


johnnykeats:
Can you explain reason of this change?

bouriquet:
Hi @johnnykeats 

bonded interfaces does not have IP specified in the inventory, but an ifcfg file should be generated

Actually, this inventory example is provided in the nic role README :

```
network_interfaces:
  eth0:
    type: bond-slave
    master: bond0
  eth1:
    type: bond-slave
    master: bond0
  bond0:
    type: bond
    vlan: false
    bond_options: "mode=4 xmit_hash_policy=layer3+4 miimon=100 lacp_rate=1"
```

but this example do not generate any ifcfg files because of this condition in the tasks :

`when: (network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none)`

oxedions:
Yes, I confirm this.

Issue #100, opened 2020-02-13
Feature request: add support for slave dns


actatux:
I was reviewing the `dns_client` and `dns_server` roles and noticed there is a maximum number of one DNS server per network.

Could we consider adding support for secondary DNS servers?
We can discuss this in this thread then I can push a PR if accepted.

oxedions:
Dear @actatux ,

This is indeed a good question.

Yes, this would be very easy to add. Maybe I would suggest that instead of:

`dns_ip: 10.10.0.7`

Two syntax would be accepted, and the Jinja code would switch from one to the other without issues:

`dns_ip: 10.10.0.7`

or

```
dns_ip:
  - 10.10.0.7
  - 10.10.0.8
```

Allowing more than 2 DNS servers...

Using an "if" with "is iterable", maybe it would worth a try.

Ben

actatux:
[Code is almost ready](https://github.com/actatux/bluebanquise/tree/feat/dns_server_slaves), sleeping since 2 months because I still need to test multiple icebergs.
Before going further, would this be implemented in dns_server role or in a new role advanced-dns_server?

oxedions:
Sorry for delay. In core, no need to go to advanced-core. DNS can be more advanced, as /etc/hosts can do the trick for basic usage.

PR #201, opened 2020-05-05
feat: dns server slaves

actatux:
PR for #100 
May be easier to review per commit.

Commit 1:

    feat: configure several DNS servers in the clients
    
    Allow to configure a list of several name servers in the dns_ip
    parameter of a network:
    
      dns_ip:
        - 10.11.0.1
        - 10.11.0.2
    
    Update the dns_client role to create as much nameserver entries in
    /etc/resolv.conf.
    
    Update the dhcp_server and advanced_dhcp_server roles to add all the
    dns_ip entries to domain-name-servers.
    
    Enable jinja2's lstrip_blocks to indent the j2 templates without
    indenting the destination file.

Commit 2:

    feat: named listen-on local ipv4 address
    
    Do not listen on all the dns_ip IPv4 addresses but only on those
    already configured on the host.

Commit 3:

    feat: add support for slave DNS servers
    
    The master DNS is the first element of the dns_ip in the primary
    management network's configuration. I assume the primary management
    network is the network of the Ansible controller.
    
      networks:
        ice1-1:
          services_ip:
            dns_ip:
              - 10.10.0.1  # Master DNS server
              - 10.10.0.2
    
    The role dns_server will create the forward and reverse zones on the
    master DNS server with the new task file master.yml. For each zone, the
    role creates an SOA file that is included in the zone file. This allows
    to update the zone serial only when the zone file changed. The serial
    number is the Unix epoch (32-bits) at time of the zone change.
    
    To add slave DNS servers, one need to configure the dns_slaves parameter
    with a list of IP addresses in the primary management network. A good
    practice would be to add this in the general_settings/network.yml
    inventory file.
    
      dns_slaves:
        - 10.10.0.2
        - 10.10.0.3

### DIFF
--- a/roles/advanced-core/advanced_dhcp_server/templates/dhcpd.networks.conf.j2
+++ b/roles/advanced-core/advanced_dhcp_server/templates/dhcpd.networks.conf.j2
@@ -1,3 +1,4 @@
+#jinja2: lstrip_blocks: True
 #### Blue Banquise file ####
 ## {{ansible_managed}}
 
@@ -20,33 +21,38 @@
 shared-network {{shared_network}} {
 {% for network in networks %}
 {% if (networks[network]['shared_network'] is defined and not none) and (networks[network]['shared_network'] == shared_network) and (networks[network]['is_in_dhcp'] == true) %}
+  {% set nameservers = networks[network]['services_ip']['dns_ip'] | ipaddr %}
 
 subnet {{networks[network]['subnet']}} netmask {{networks[network]['netmask']}} {
 
-{% if networks[network]['dhcp_unknown_range'] is defined and not none %}
+  {% if networks[network]['dhcp_unknown_range'] is defined and not none %}
   range {{networks[network]['dhcp_unknown_range']}};
-{% endif %}
+  {% endif %}
   option domain-name "{{domain_name}}";
-  option domain-name-servers {{networks[network]['services_ip']['dns_ip']}};
+  {% if nameservers is string %}
+  option domain-name-servers {{nameservers}};
+  {% else %}
+  option domain-name-servers {{nameservers | join(', ')}};
+  {% endif %}
   option broadcast-address {{networks[network]['broadcast']}};
   option ntp-servers {{networks[network]['services_ip']['time_ip']}};
-{% if networks[network]['gateway'] is defined and not none %}
+  {% if networks[network]['gateway'] is defined and not none %}
   option routers {{networks[network]['gateway']}};
-{% endif %}
+  {% endif %}
   default-lease-time {{dhcp_settings['default_lease_time']}};
   max-lease-time {{dhcp_settings['max_lease_time']}};
-{% if networks[network]['services_ip']['pxe_ip'] is defined and not none %}
-{% if (networks[network]['services_ip']['pxe_http_ip'] is defined and not none)
-    and (networks[network]['services_ip']['pxe_http_ip'] != networks[network]['services_ip']['pxe_ip']) %}
+  {% if networks[network]['services_ip']['pxe_ip'] is defined and not none %}
+    {% if (networks[network]['services_ip']['pxe_http_ip'] is defined and not none)
+      and (networks[network]['services_ip']['pxe_http_ip'] != networks[network]['services_ip']['pxe_ip']) %}
   if exists user-class and option user-class = "BlueBanquise" {
     next-server {{networks[network]['services_ip']['pxe_http_ip']}};
   } else {
     next-server {{networks[network]['services_ip']['pxe_ip']}};
   }
-{% else %}
+    {% else %}
   next-server {{networks[network]['services_ip']['pxe_ip']}};
-{% endif %}
-{% endif %}
+    {% endif %}
+  {% endif %}
 
   include "/etc/dhcp/dhcpd.{{network}}.conf";
 
@@ -61,31 +67,38 @@ subnet {{networks[network]['subnet']}} netmask {{networks[network]['netmask']}} 
 {% for network in networks %}
 {% if j2_current_iceberg_network in network %}
 {% if (networks[network]['shared_network'] is not defined or none) and (networks[network]['is_in_dhcp'] == true) %}
+  {% set nameservers = networks[network]['services_ip']['dns_ip'] | ipaddr %}
 
 subnet {{networks[network]['subnet']}} netmask {{networks[network]['netmask']}} {
 
-  {% if networks[network]['dhcp_unknown_range'] is defined and not none %}range {{networks[network]['dhcp_unknown_range']}};{% endif %}
-
+  {% if networks[network]['dhcp_unknown_range'] is defined and not none %}
+  range {{networks[network]['dhcp_unknown_range']}};
+  {% endif %}
   option domain-name "{{domain_name}}";
-  option domain-name-servers {{networks[network]['services_ip']['dns_ip']}};
+  {% if nameservers is string %}
+  option domain-name-servers {{nameservers}};
+  {% else %}
+  option domain-name-servers {{nameservers | join(', ')}};
+  {% endif %}
   option broadcast-address {{networks[network]['broadcast']}};
   option ntp-servers {{networks[network]['services_ip']['time_ip']}};
-  {% if networks[network]['gateway'] is defined and not none %}option routers {{networks[network]['gateway']}};{% endif %}
-
+  {% if networks[network]['gateway'] is defined and not none %}
+  option routers {{networks[network]['gateway']}};
+  {% endif %}
   default-lease-time {{dhcp_settings['default_lease_time']}};
   max-lease-time {{dhcp_settings['max_lease_time']}};
-{% if networks[network]['services_ip']['pxe_ip'] is defined and not none %}
-{% if (networks[network]['services_ip']['pxe_http_ip'] is defined and not none)
-    and (networks[network]['services_ip']['pxe_http_ip'] != networks[network]['services_ip']['pxe_ip']) %}
+  {% if networks[network]['services_ip']['pxe_ip'] is defined and not none %}
+    {% if (networks[network]['services_ip']['pxe_http_ip'] is defined and not none)
+      and (networks[network]['services_ip']['pxe_http_ip'] != networks[network]['services_ip']['pxe_ip']) %}
   if exists user-class and option user-class = "BlueBanquise" {
     next-server {{networks[network]['services_ip']['pxe_http_ip']}};
   } else {
     next-server {{networks[network]['services_ip']['pxe_ip']}};
   }
-{% else %}
+    {% else %}
   next-server {{networks[network]['services_ip']['pxe_ip']}};
-{% endif %}
-{% endif %}
+    {% endif %}
+  {% endif %}
 
   include "/etc/dhcp/dhcpd.{{network}}.conf";
 
--- a/roles/core/dns_client/templates/resolv.conf.j2
+++ b/roles/core/dns_client/templates/resolv.conf.j2
@@ -1,8 +1,16 @@
+#jinja2: lstrip_blocks: True
 #### Blue Banquise file ####
 ## {{ansible_managed}}
+{% set nameservers = networks[j2_node_main_network]['services_ip']['dns_ip'] | ipaddr %}
 
 search {{domain_name}}
-nameserver {{networks[j2_node_main_network]['services_ip']['dns_ip']}} 
+{% if nameservers is string %}
+nameserver {{nameservers}}
+{% else %}
+  {% for nameserver in nameservers %}
+nameserver {{nameserver}}
+  {% endfor %}
+{% endif %}
 {% if external_dns is defined and external_dns is not none %}
 {% if external_dns.dns_client is defined and external_dns.dns_client is not none and external_dns.dns_client %}
 {% for external_server in external_dns.dns_client %}
--- a/roles/core/dns_server/tasks/main.yml
+++ b/roles/core/dns_server/tasks/main.yml
@@ -26,6 +26,31 @@
   tags:
     - package
 
+# Master DNS is the dns_ip of primary management network
+# (management_network_naming + '1-1')
+- name: Set master DNS server - Get DNS IP
+  set_fact:
+    dns_master: "{{ networks[management_network_naming+'1-1'].services_ip.dns_ip }}"
+  when:
+    - "dns_master is not defined"
+    - "networks[j2_node_main_network].services_ip.dns_ip is defined"
+
+- name: Assert dns_master is defined
+  assert:
+    that:
+      - dns_master is defined
+    fail_msg: "'dns_master' must be defined. Please ensure that either 'dns_master' or
+              networks[{{ management_network_naming }}1-1].services_ip.dns_ip is
+              defined and contains a valid IP address."
+
+# If dns_ip is a list, pick the first value as master DNS
+- name: Set master DNS server - Pick first element of dns_ip
+  set_fact:
+    dns_master: "{{ dns_master | first }}"
+  when:
+    - "dns_master is iterable"
+    - "dns_master is not string"
+
 - name: Create needed directory >> /var/named
   file:
     path: /var/named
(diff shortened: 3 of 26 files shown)


actatux:
Test is failing because the ipaddr filter requires python's netaddr be installed on the ansible controller.
I talked about this a couple of month ago in the chat. Should I add the package to the core/ansible role?

roles/advanced-core/advanced_dhcp_server/templates/dhcpd.networks.conf.j2
```diff
@@ -20,33 +21,38 @@
...
+  {% if networks[network]['dhcp_unknown_range'] is defined and not none %}
   range {{networks[network]['dhcp_unknown_range']}};
-{% endif %}
+  {% endif %}
   option domain-name "{{domain_name}}";
-  option domain-name-servers {{networks[network]['services_ip']['dns_ip']}};
+  {% if nameservers is string %}
+  option domain-name-servers {{nameservers}};
+  {% else %}
+  option domain-name-servers {{nameservers | join(', ')}};
```
oxedions, line 35: It is nice you kept compatibility with string way. :-)

roles/core/dns_client/templates/resolv.conf.j2
```diff
@@ -1,8 +1,16 @@
+#jinja2: lstrip_blocks: True
 #### Blue Banquise file ####
 ## {{ansible_managed}}
+{% set nameservers = networks[j2_node_main_network]['services_ip']['dns_ip'] | ipaddr %}
```
oxedions, line 4: I like the idea of filtering ip, to ensure no issues.

roles/core/dns_server/tasks/main.yml
```diff
@@ -26,6 +26,27 @@
...
+    dns_master: "{{ networks[management_network_naming+'1-1'].services_ip.dns_ip }}"
+  when:
+    - "dns_master is not defined"
+    - "networks[j2_node_main_network].services_ip.dns_ip is defined"
+
+- name: Assert dns_master is defined
+  assert:
+    that:
+      - dns_master is defined
+    fail_msg: "'dns_master' must be defined"
```
oxedions, line 42: I think this message is not verbose enough. Wouldn't it be better to say :
`"Could not get a dns_master. Please ensure that dns_master variable is defined, or that networks[{{j2_node_main_network}}].services_ip.dns_ip is defined and contains a valid ip or list of ip."`

roles/core/dns_server/tasks/main.yml
```diff
@@ -43,25 +64,9 @@
...
-  template:
-    src: reverse.j2
-    dest: /var/named/reverse
-    owner: root
-    group: root
-    mode: 0644
-  notify: Restart dns services
+- name: Configure Master DNS
+  include_tasks: master.yml
+  when: "dns_master in ansible_all_ipv4_addresses"
```
oxedions, line 69: After spending some time computing on this, I don't like the usage of ansible_all_ipv4_addresses here. I will explain in a next comment in detail.

roles/core/dns_server/tasks/master.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: Template >> Forward zone file on Master DNS
+  template:
+    src: forward.j2
+    dest: /var/named/forward
+    owner: root
+    group: root
+    mode: 0644
+  notify: Restart dns services
+  register: zoneconf
+  when: "dns_master in ansible_all_ipv4_addresses"
```
oxedions, line 11: The file `roles/core/dns_server/tasks/master.yml` is executed only if "dns_master in ansible_all_ipv4_addresses" in  `roles/core/dns_server/tasks/main.yml`.
Isn't this condition redundant ?

roles/core/dns_server/tasks/master.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: Template >> Reverse zone file on Master DNS
+  template:
+    src: reverse.j2
+    dest: /var/named/reverse
+    owner: root
+    group: root
+    mode: 0644
+  notify: Restart dns services
+  register: zoneconf
+  when: "dns_master in ansible_all_ipv4_addresses"
```
oxedions, line 43: Same here. Redundant ?

roles/core/dns_server/templates/named.conf.j2
```diff
@@ -1,13 +1,24 @@
...
 options {
   listen-on port 53 {
     127.0.0.1;
 {% for network in (networks | select('match',(j2_current_iceberg_network+'-[0-9]+')) | list | unique | sort) %}
-{% if networks[network]['is_in_dns'] is defined and networks[network]['is_in_dns'] == true %}
-    {{networks[network]['services_ip']['dns_ip']}};
-{% endif %}
+  {% if networks[network]['is_in_dns'] is defined and networks[network]['is_in_dns'] == true %}
+    {% set addresses = networks[network]['services_ip']['dns_ip'] | ipaddr %}
+    {% if dns_slaves is defined %}
```
oxedions, line 11: There are no register to define dns_slaves according to `networks[j2_node_main_network].services_ip.dns_ip`, i.e. not first ones. If I understand, first in list is master (if note defined manually), and others are slaves. Is it expected ?

roles/core/dns_server/templates/named.conf.j2
```diff
@@ -1,13 +1,24 @@
...
+      {% set addresses = addresses|union(dns_slaves) %}
+    {% endif %}
+    {% if addresses is string and addresses in ansible_all_ipv4_addresses %}
```
oxedions, line 14: Let me explain here in details what makes me not confident about this.

I understand your target here : checking if ip exist on current host, and if yes, add them in the listen part.

However, I would prefer to use the hostvars['network_interfaces'] part. It may appends that for any reason, an interface is down, or an ip is not set on the dns node. Using hostvars['network_interfaces'] would make the named service fail to start as all expected interfaces are not up, and force admin to check what is going on.

If we rely on facts like ansible_all_ipv4_addresses, configuration is dynamically generated, but this also means that configuration deployed could not be always the expected one.
For example an admin deploying dns before configuring all the network, then start service, and configure network after that. dns will not listen on all expected interfaces.

Not sure it is clear. I often prefer a service that fails vs a service that __may__ not provide expected features (like I prefer a nice segmentation fault in C vs a code that work but may not provide the good result).

roles/core/dns_client/templates/resolv.conf.j2
actatux: See my previous comment: the ipaddr filter requires python's netaddr be installed on the ansible controller.
So if we want to keep this filter, we need to install the package on the ansible controller. Best option would be the ansible role.

roles/core/dns_server/templates/named.conf.j2
actatux: dns_slaves can be defined in the general_settings/network.yml inventory file.
This is the list of IP addresses of dns slaves connected to the main network of the master.

  dns_slaves:
    - 10.10.0.2
    - 10.10.0.3

This ensures the master can notify the slaves when there is an update in a DNS map. If dnf_slaves is not defined, the dns master won't know which nodes it has to notify, which will delay the update of the maps on the slaves.

This is useful in multi icebergs where the main network of the master can be ice1-1 while the main network of some slaves is ice2-1. Nodes in ice2-1 will be able to use `networks[j2_node_main_network].services_ip.dns_ip` to request the dns servers (slaves), but the master is not connected to this network.

roles/core/dns_server/templates/named.conf.j2
actatux: I get your point. Let's wait for additional reviews/comments.

> If we rely on facts like ansible_all_ipv4_addresses, configuration is dynamically generated, but this also means that configuration deployed could not be always the expected one.

This could be the expected configuration. I want to start the dns server on configured interfaces. Some interfaces could be unconfigured on purpose, why should we prevent to start dns?

> For example an admin deploying dns before configuring all the network, then start service, and configure network after that. dns will not listen on all expected interfaces.

The dns server will not listen on all interfaces defined in the inventory until the next run of the playbook.
We might see things differently: I assume the admin knows what he is doing with the system. We do provide example playbooks where network configuration is almost at the beginning of the play.
I don't believe the stack should handle the corner case where the admin configures network services first, then network, then never run the playbook again to ensure the configuration is correct. IMHO, any change to the network configuration should be followed by a run of the playbook on the host to ensure everything is configured as expected.

roles/core/dns_server/tasks/master.yml
actatux: Yes. I'll double check.

actatux: Good catch. At first these tasks were in the main.yml, then I moved them to the master.yml but forgot to remove the when clause. I'll update the code.

roles/core/dns_server/templates/named.conf.j2
actatux: >  If I understand, first in list is master (if note defined manually), and others are slaves. Is it expected ?

Indeed, if `dns_master` is not defined, first element of the list `networks[j2_node_main_network].services_ip.dns_ip` is the master when j2_node_main_network is the primary management network (management_network_naming + '1-1'). Other elements in the list are slaves dns servers available in this network. They are used to render resolv.conf and dhcp configuration (nameservers).

In multi icebergs, an inventory looks like:
```
  inventory:
    vars:
      networks:
        ice1-1:
          subnet: 10.11.0.0
          prefix: 16
          netmask: 255.255.0.0
          broadcast: 10.11.255.255
          is_in_dns: true
          services_ip:
            dns_ip:
              - 10.11.0.1
              - 10.11.0.2
        ice2-1:
          subnet: 10.12.0.0
          prefix: 16
          netmask: 255.255.0.0
          broadcast: 10.12.255.255
          is_in_dns: true
          services_ip:
            dns_ip:
              - 10.12.0.1
              - 10.12.0.2
      external_dns: []
      management_network_naming: "ice"
      domain_name: example.com
      dns_slaves:
        - 10.11.0.2
        - 10.11.0.3
        - 10.11.0.4
      start_services: true
      enable_services: true
    host_vars:
      mgmt0-master:
        j2_current_iceberg_network: ice1
        j2_node_main_network: "ice1-1"
        j2_node_main_network_interface: en0
        network_interfaces:
          en0:
            ip4: 10.11.0.1
            network: ice1-1
      mgmt1-slave:
        j2_current_iceberg_network: ice1
        j2_node_main_network: "ice1-1"
        j2_node_main_network_interface: en0
        network_interfaces:
          en0:
            ip4: 10.11.0.2
            network: ice1-1
      mgmt2-slave:
        j2_current_iceberg_network: ice2
        j2_node_main_network: "ice2-1"
        j2_node_main_network_interface: en1
        network_interfaces:
          en0:
            ip4: 10.11.0.3
            network: ice1-1
          en1:
            ip4: 10.12.0.1
            network: ice2-1
      mgmt3-slave:
        j2_current_iceberg_network: ice2
        j2_node_main_network: "ice2-1"
        j2_node_main_network_interface: en1
        network_interfaces:
          en0:
            ip4: 10.11.0.4
            network: ice1-1
          en1:
            ip4: 10.12.0.2
            network: ice2-1
```

mgmt0-master would be the master DNS server handling the zone files for domain example.com. All other servers will be slaves.
Hosts in network ice1-1 will use nameservers 10.11.0.1 and 10.11.0.2 (`networks["ice1-1"].services_ip.dns_ip`), while hosts in network ice2-1 will use nameservers 10.12.0.1 and 10.12.0.2 (`networks["ice2-1"].services_ip.dns_ip`).

For any update in the DNS map file, the master will notify dns_slaves (10.11.0.[2-4]). The slaves will then start a zone transfer (AXFR) to update their local copy.

actatux: While writing the code, I also thought using `networks[j2_node_main_network].services_ip.dns_master` and `networks[j2_node_main_network].services_ip.dns_slaves` instead of global dns_master and dns_slaves.
However I find it more error-prone because it might led to think the stack can handle several master/slaves configuration.

roles/core/dns_server/tasks/main.yml
actatux: Is this better?
```
fatal: [mgmt0-master]: FAILED! => {
    "assertion": "dns_master is defined", 
    "changed": false, 
    "evaluated_to": false, 
    "msg": "'dns_master' must be defined. Please ensure that either 'dns_master' or networks[ice1-1].services_ip.dns_ip is defined and contains a valid IP address."
}
```

roles/core/dns_client/templates/resolv.conf.j2
johnnykeats: OK for netaddr. Useful.

roles/core/dns_server/templates/named.conf.j2
johnnykeats: `if ( dns_slaves is not defined ) and ( nameservers is not string )` -> dns_slave is ignored ?

to confirm, not a DNS master

roles/core/dns_server/templates/named.conf.j2
johnnykeats: Got the two points

Issue as already been seen and discussed in other roles.

Stack is design to be didactic and 'stupid' for new system administrators, while also trying to be 'smart' and dynamic for advanced users.

Would propose this, not only for this role, but for future:

A variable in inventory, called `local_reference_data`, by default to `facts`, but that could be `inventory`.
Note: facts means all local ways to gather data. May be facts, maybe task command to register, maybe target tool/service dynamic ways to gather data.

All Core and Advanced-Core roles must comply with this variable. Not now, but on long term.

Rule is not mandatory for addons. Addon developer do it the way desired.

Light code overhead

roles/core/dns_server/templates/named.conf.j2
actatux: If `dns_slaves is not defined` but `nameservers is a list`, the dns slaves won't be ignored. However, the dns master won't know which IP addresses to notify in network1-1, and the slaves in networkX-1 (X>1) won't listen on the interface connected to network1-1. The slaves will refresh their configuration every 3600 seconds (based on refresh value in the SOA), or when the admin will run the `rndc reload $domain` command.
The `dns_slaves` is needed for master->slaves communication only.

roles/core/dns_server/templates/named.conf.j2
actatux: OK. For this PR I will revert to inventory instead of facts, as per @oxedions request. Then we can implement the `local_reference_data` in a dedicated PR.

roles/core/dns_server/tasks/main.yml
actatux: Done

roles/core/dns_server/tasks/main.yml
actatux: Done

roles/core/dns_server/templates/named.conf.j2
actatux: Done

roles/core/dns_client/templates/resolv.conf.j2
actatux: Done.

oxedions: Thanks ! 👍

roles/core/dns_server/templates/named.conf.j2
oxedions: I opened https://github.com/bluebanquise/bluebanquise/issues/207, so we do not forget.

roles/core/dns_server/templates/named.conf.j2
oxedions: Ok for me. Thanks for explaining all of this :-)

actatux:
@oxedions I had to push a new commit to install netaddr and jmespath to pass CI tests for dns_server and dns_client. It dismissed your previous review.

oxedions:
Hello @actatux 

Very low work to do compared to the one you made for this PR :-)

Many thanks !

oxedions:
One @johnnykeats validates, do you want to Squash when merging or not ?

actatux:
Hello, simple merge please, no squash. I like to keep commits separated by topics.

actatux:
@oxedions I changed roles versions in the last commit.

PR #202, opened 2020-05-05
Fix: remove clustershell example groups

oxedions:
Fix clustershell role, by using local.cfg directly instead of adding another file.


oxedions:
I removed the file.

oxedions:
@johnnykeats ?

PR #209, opened 2020-05-07
Fix: solve VLAN issues

oxedions:
missing good TYPE, missing VLAN_ID, missing PHYSDEV

### DIFF
--- a/roles/core/nic/tasks/main.yml
+++ b/roles/core/nic/tasks/main.yml
@@ -23,5 +23,5 @@
     mode: 0644
   with_items: "{{ network_interfaces }}"
   when: >
-      ( network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none ) or 
-      ( network_interfaces[item]['type'] is defined and network_interfaces[item]['type'] is not none and network_interfaces[item]['type'] == 'bond-slave')
+      ( network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none ) or
+      ( network_interfaces[item]['type'] is defined and network_interfaces[item]['type'] is not none and network_interfaces[item]['type'] == 'bond-slave' )
--- a/roles/core/nic/tasks/main.yml
+++ b/roles/core/nic/tasks/main.yml
@@ -22,4 +22,6 @@
     group: root
     mode: 0644
   with_items: "{{ network_interfaces }}"
-  when: (network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none) or ((network_interfaces[item]['type'] is defined and not none) and (network_interfaces[item]['type'] == 'bond-slave'))
+  when: >
+      ( network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none ) or 
+      ( network_interfaces[item]['type'] is defined and network_interfaces[item]['type'] is not none and network_interfaces[item]['type'] == 'bond-slave')
--- a/roles/core/nic/templates/ifcfg.j2
+++ b/roles/core/nic/templates/ifcfg.j2
@@ -21,6 +21,9 @@ BOOTPROTO=none
 {% set nic_type = 'Infiniband' %}
 {% endif %}
 {% endif %}
+{% if (network_interfaces[item]['vlan'] is defined) and (network_interfaces[item]['vlan'] == true) %}
+{% set nic_type = 'Vlan' %}
+{% endif %}
 TYPE={{nic_type}}
 
 {# 3. Now basic network parameters, if exist. #}
@@ -67,6 +70,8 @@ MTU={{network_interfaces[item]['mtu']}}
 {# 6. Is it a VLAN ? #}
 {% if (network_interfaces[item]['vlan'] is defined and not none) and (network_interfaces[item]['vlan'] == true) %}
 VLAN=yes
+PHYSDEV={{network_interfaces[item]['physical_device']}}
+VLAN_ID={{network_interfaces[item]['vlan_id']}}
 {% endif %}
 
 {# 7. Is it a bonding (bond) or a bonding-slave (bond-slave) ? #}
@@ -82,4 +87,3 @@ BONDING_OPTS={{network_interfaces[item]['bond_options']}}
 SLAVE=yes
 MASTER={{network_interfaces[item]['master']}}
 {% endif %}
-
--- a/roles/core/nic/templates/ifcfg.j2
+++ b/roles/core/nic/templates/ifcfg.j2
@@ -1,56 +1,59 @@
+#jinja2: lstrip_blocks: "True"
 #### Blue Banquise file ####
 ## {{ansible_managed}}
 
 {# 1. First the name and default parameters. #}
 NAME={{item}}
 DEVICE={{item}}
-{% if (network_interfaces[item]['autoconnect'] is defined and not none) and (network_interfaces[item]['autoconnect'] == false) %}
+{% if (network_interfaces[item]['autoconnect'] is defined and network_interfaces[item]['autoconnect'] is not none) and (network_interfaces[item]['autoconnect'] == false) %}
 ONBOOT=no
 {% else %}
 ONBOOT=yes
 {% endif %}
+{% if ansible_facts.distribution_major_version < 8 %}
 NM_CONTROLLED={{nic_nm_controlled | ternary('yes','no')}}
+{% endif %}
 BOOTPROTO=none
 
 {# 2. Let's get the type. Default is Ethernet. #}
 {% set nic_type = 'Ethernet' %}
 {% if network_interfaces[item]['type'] is defined and not none %}
-{% if network_interfaces[item]['type'] == 'bond' %}
-{% set nic_type = 'Bond' %}
-{% elif network_interfaces[item]['type'] == 'infiniband' %}
-{% set nic_type = 'Infiniband' %}
-{% endif %}
+  {% if network_interfaces[item]['type'] == 'bond' %}
+    {% set nic_type = 'Bond' %}
+  {% elif network_interfaces[item]['type'] == 'infiniband' %}
+    {% set nic_type = 'Infiniband' %}
+  {% endif %}
 {% endif %}
 {% if (network_interfaces[item]['vlan'] is defined) and (network_interfaces[item]['vlan'] == true) %}
-{% set nic_type = 'Vlan' %}
+  {% set nic_type = 'Vlan' %}
 {% endif %}
 TYPE={{nic_type}}
 
 {# 3. Now basic network parameters, if exist. #}
 {# 3.1 Lets check if we need to set a simple ip, or multiple. #}
-{% if network_interfaces[item]['ip4'] is defined and not none %}
-{# 3.1.1 Ok, now lets check we have a prefix or a netmask for this ip. Can come from network or directly set in network_interfaces. #}
-{% if (network_interfaces[item]['network'] is defined and not none) and (networks[network_interfaces[item]['network']] is defined and not none) %}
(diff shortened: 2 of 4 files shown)


johnnykeats:
Agree with @actatux. Splitting is perfect.

actatux:
Proposal:
```
diff --git a/roles/core/nic/templates/ifcfg.j2 b/roles/core/nic/templates/ifcfg.j2
index ac5670a..7af4549 100644
--- a/roles/core/nic/templates/ifcfg.j2
+++ b/roles/core/nic/templates/ifcfg.j2
@@ -70,9 +70,15 @@ MTU={{network_interfaces[item]['mtu']}}
 {# 6. Is it a VLAN ? #}
 {% if (network_interfaces[item]['vlan'] is defined and not none) and (network_interfaces[item]['vlan'] == true) %}
 VLAN=yes
+{% set vlan_itf = item.split('.') %}
+{% if vlan_itf | length == 2 %}
+PHYSDEV={{vlan_itf[0]}}
+VLAN_ID={{vlan_itf[1]}}
+{% else %}
 PHYSDEV={{network_interfaces[item]['physical_device']}}
 VLAN_ID={{network_interfaces[item]['vlan_id']}}
 {% endif %}
+{% endif %}
 
 {# 7. Is it a bonding (bond) or a bonding-slave (bond-slave) ? #}
 {# 7.1 Master first. #}

```

actatux:
While not directly related to this change, I notice that BOOTPROTO=none but there is no static configuration.

From /usr/share/doc/initscripts/sysconfig.txt (CentOS 8.1):
```
    BOOTPROTO=none|bootp|dhcp
      'bootp' or 'dhcp' cause a DHCP client to run on the device. Any other
      value causes any static configuration in the file to be applied.
```
Here is the file created by the role:
```
# grep -v "^$" /etc/sysconfig/network-scripts/ifcfg-vlan100
#### Blue Banquise file ####
## Ansible managed: modified on 2020-05-07 20:08:52 by root on management1
NAME=vlan100
DEVICE=vlan100
ONBOOT=yes
NM_CONTROLLED=yes
BOOTPROTO=none
TYPE=Vlan
VLAN=yes
PHYSDEV=eth2
VLAN_ID=100
```

oxedions:
Hello.

Regarding this last point, this is a bug. When `type` is defined, there should be `ip4` value if type = vlan. I will correct it.

oxedions:
Regarding the proposal of dividing name, I agree with both of you for bug fix in 1.2 branchs.

However, I am not fan of having a syntax with no PHYSDEV and VLAN_ID defined, as it is much cleaner to have these values defined in the inventory.

I will fix it right now with your proposal, so we can backport in previous branchs, but then I will introduce a breaking change in another PR for the 1.3, and force usage of these two values :-)

actatux:
Otherwise we merge the code as is for master and we can update the backport to avoid breaking change in the minor?

oxedions:
Let's do it this way. I will correct the issue for ipv4, and then we merge this PR into master, and I will open another PR as backport in 1.2 stable branch, using your split method.

Then, when adding OpenSuse support, I will add Os gather like the other roles.

oxedions:
> # grep -v "^$" /etc/sysconfig/network-scripts/ifcfg-vlan100
> #### Blue Banquise file ####
> ## Ansible managed: modified on 2020-05-07 20:08:52 by root on management1
> NAME=vlan100
> DEVICE=vlan100
> ONBOOT=yes
> NM_CONTROLLED=yes
> BOOTPROTO=none
> TYPE=Vlan
> VLAN=yes
> PHYSDEV=eth2
> VLAN_ID=100

@actatux Can you provide me inventory that generated this ? I mean the host network_interfaces hash.

roles/core/nic/templates/ifcfg.j2
```diff
@@ -1,53 +1,59 @@
...
+  {% if network_interfaces[item]['type'] == 'bond' %}
+    {% set nic_type = 'Bond' %}
+  {% elif network_interfaces[item]['type'] == 'infiniband' %}
+    {% set nic_type = 'Infiniband' %}
+  {% endif %}
 {% endif %}
+{% if (network_interfaces[item]['vlan'] is defined) and (network_interfaces[item]['vlan'] == true) %}
+  {% set nic_type = 'Vlan' %}
 {% endif %}
 TYPE={{nic_type}}
```
bouriquet, line 30: Hi @oxedions,

Proposal :

```
{% if (network_interfaces[item]['type'] is defined) %}
  {% if (network_interfaces[item]['type'] | lower) in ['infiniband', 'bond', 'vlan'] %}
TYPE={{ network_interfaces[item]['type'] | lower }}
  {% endif %}
{% else %}
TYPE=ethernet
{% endif %}
```

"and not none" instruction on this line isn't taken into account (must be "and var is not none") and not necessary, you fill TYPE with ethernet even if this variable is "none"

```
{% if network_interfaces[item]['type'] is defined and not none %}
```

You also make a difference between 'vlan' type and the others types of nic. Having a 'vlan' var in your network_interfaces dict is a little disturbing (sounds like we must fill this var with a VLAN number)

Be careful, I'm not sure type is case insensitive, only tested on RH8 and "vlan" type.

actatux:
This is the example I added in the readme (e873492):
```
network_interfaces:
  vlan100:
    vlan: true
    vlan_id: 100
    physical_device: eth2
    ip4: 10.100.0.1
    network: net-100
```

roles/core/nic/templates/ifcfg.j2
actatux: The `vlan` var in the network_interfaces matches the `VLAN=yes|no` var from sysconfig.

actatux:
Static code analysis fails:
```
201 Trailing whitespace
/home/runner/work/bluebanquise/bluebanquise/roles/core/nic/tasks/main.yml:26
      ( network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none ) or 
```

roles/core/nic/templates/ifcfg.j2
```diff
@@ -1,53 +1,59 @@
...
 {# 1. First the name and default parameters. #}
 NAME={{item}}
 DEVICE={{item}}
-{% if (network_interfaces[item]['autoconnect'] is defined and not none) and (network_interfaces[item]['autoconnect'] == false) %}
+{% if (network_interfaces[item]['autoconnect'] is defined and network_interfaces[item]['autoconnect'] is not none) and (network_interfaces[item]['autoconnect'] == false) %}
 ONBOOT=no
 {% else %}
 ONBOOT=yes
 {% endif %}
+{% if ansible_facts.distribution_major_version < 8 %}
```
actatux, line 13:
```suggestion
{% if ansible_facts.distribution_major_version|int < 8 %}
```

Otherwise templating fails with: '<' not supported between instances of 'AnsibleUnsafeText' and 'int'

roles/core/nic/templates/ifcfg.j2
```diff
@@ -1,53 +1,59 @@
...
+{% if network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none %}
+  {# 3.1.1 Ok, now lets check we have a prefix or a netmask for this ip. Can come from network or directly set in network_interfaces. #}
+  {% if (network_interfaces[item]['network'] is defined and not none) and (networks[network_interfaces[item]['network']] is defined and not none) %}
+    {% if networks[network_interfaces[item]['network']]['prefix'] is defined and not none %}
+      {# 3.1.2 Ok, go for prefix from network. #}
 IPADDR={{network_interfaces[item]['ip4']}}
 PREFIX={{networks[network_interfaces[item]['network']]['prefix']}}
-{% elif  networks[network_interfaces[item]['network']]['netmask'] is defined and not none %}
-{# 3.1.3 Ok, go for netmask from network. #}
+    {% elif  networks[network_interfaces[item]['network']]['netmask'] is defined and not none %}
```
actatux, line 41: Can you fix "and not none"?

roles/core/nic/templates/ifcfg.j2
```diff
@@ -67,19 +73,20 @@ MTU={{network_interfaces[item]['mtu']}}
...
+PHYSDEV={{network_interfaces[item]['physical_device']}}
+VLAN_ID={{network_interfaces[item]['vlan_id']}}
 {% endif %}
 
 {# 7. Is it a bonding (bond) or a bonding-slave (bond-slave) ? #}
 {# 7.1 Master first. #}
 {% if (network_interfaces[item]['type'] is defined and not none) and (network_interfaces[item]['type'] == 'bond') %}
 BONDING_MASTER=yes
-{% if network_interfaces[item]['bond_options'] is defined and not none %}
+  {% if network_interfaces[item]['bond_options'] is defined and not none %}
```
actatux, line 84: Can you fix "and not none"?

roles/core/nic/tasks/main.yml
```diff
@@ -22,4 +22,6 @@
     group: root
     mode: 0644
   with_items: "{{ network_interfaces }}"
-  when: (network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none) or ((network_interfaces[item]['type'] is defined and not none) and (network_interfaces[item]['type'] == 'bond-slave'))
+  when: >
+      ( network_interfaces[item]['ip4'] is defined and network_interfaces[item]['ip4'] is not none ) or
```
actatux, line 26: 201 Trailing whitespace

hmeAtos:
> Proposal:
> 
> ```
> diff --git a/roles/core/nic/templates/ifcfg.j2 b/roles/core/nic/templates/ifcfg.j2
> index ac5670a..7af4549 100644
> --- a/roles/core/nic/templates/ifcfg.j2
> +++ b/roles/core/nic/templates/ifcfg.j2
> @@ -70,9 +70,15 @@ MTU={{network_interfaces[item]['mtu']}}
>  {# 6. Is it a VLAN ? #}
>  {% if (network_interfaces[item]['vlan'] is defined and not none) and (network_interfaces[item]['vlan'] == true) %}
>  VLAN=yes
> +{% set vlan_itf = item.split('.') %}
> +{% if vlan_itf | length == 2 %}
> +PHYSDEV={{vlan_itf[0]}}
> +VLAN_ID={{vlan_itf[1]}}
> +{% else %}
>  PHYSDEV={{network_interfaces[item]['physical_device']}}
>  VLAN_ID={{network_interfaces[item]['vlan_id']}}
>  {% endif %}
> +{% endif %}
>  
>  {# 7. Is it a bonding (bond) or a bonding-slave (bond-slave) ? #}
>  {# 7.1 Master first. #}
> ```

I add your proposal in 1.1.2 branch.

roles/core/nic/templates/ifcfg.j2
oxedions: I tried to match sysconfig strictly. But I also understand your concern.

Question is: can it happen (not sure) that a configuration is using type != vlan but VLAN=yes ?

oxedions:
Bonding is also buggy.

Will commit another version.

roles/core/nic/templates/ifcfg.j2
actatux: It seems not. I only find [one place in the code](https://github.com/NetworkManager/NetworkManager/blob/master/src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-writer.c#L1291) where NetworkManager writes VLAN=yes in the ifcfg file. This is the master branch, I did not check the sources of EL7/8.

actatux: https://github.com/NetworkManager/NetworkManager/commit/e8ffca398f429e2d8ac4e2131c9aa54d13fb3fbf

actatux: The NM ifcfg reader condition is not case sensitive: `(!strcasecmp (type, TYPE_VLAN))`

Issue #205, opened 2020-05-05
Issue with OS variables gathering and variable precedence


actatux:
Hello,

I started to change the way to gather OS variables in roles with #173 and #179, then put the work on hold:
> Comment on #173 fix: use with_first_found to gather OS vars
> By the way @actatux  
> I would like to not update more roles with this enhancement.  
> I would like to see how users react to this now :-)

We currently have some variable precedence issues because of these changes.

When running a playbook with a debug task, we can see the following:
```
# /usr/bin/ansible-playbook /etc/bluebanquise/playbooks/management1.yml \
    --extra-vars target=management1 --tags dns_server,time
[...]
TASK [time : DBG - Display value of 'packages_to_install'] ****************************
ok: [management1] => 
  msg:
  - bind

TASK [time : Install packages] ********************************************************
fatal: [management1]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable.
    The error was: 'list object' has no attribute 'centos'
```
The role time fails to install the packages. `packages_to_install = bind` because it was defined in the playbook scope by include_vars in the dns_server role, even if `packages_to_install` is also defined in time/vars/main.yml.

Obviously if we run the role individually, there is no issue:
```
# /usr/bin/ansible-playbook /etc/bluebanquise/playbooks/management1.yml \
    --extra-vars target=management1 --tags time
[...]
TASK [time : DBG - Display value of 'packages_to_install'] ****************************
ok: [management1] => 
  msg:
    centos:
      _7:
      - chrony
      _8:
      - chrony
    redhat:
      _7:
      - chrony
      _8:
      - chrony
    ubuntu:
      _18:
      - chrony
      - tzdata
```

 Extract of [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence):

> Here is the order of precedence from least to greatest (the last listed variables winning prioritization):
> [...]
> 15. role vars (defined in role/vars/main.yml)
> [...]
> 18. include_vars
> [...]

We need to decide if we move forward or revert the changes before next release. An option could be to prefix the variables with role names (eg. dns_server_packages_to_install).

johnnykeats:
Added

```
# by default, variables from roles will be visible in the global variable
# scope. To prevent this, the following option can be enabled, and only
# tasks and handlers within the role will see the variables there
private_role_vars = yes
```

in ansible.cfg to prevent this issue before.

Can you check your variable?
[...]

PR #217, opened 2020-05-12
fix: gather OS vars in core roles

actatux:
Search vars in:
 - `<distribution>_<major>.yml`    # eg. CentOS_8.yml
 - `<os_family>_<major>.yml`       # eg. RedHat_8.yml
 - `<distribution>.yml`            # eg. CentOS.yml
 - `<os_family>.yml`               # eg. RedHat.yml

Prefix role variables with the role name.

Port code to Ansible fact namespacing: ansible_* -> ansible_facts.*

List of roles:
 - advanced_dhcp_server
 - ansible
 - conman
 - dhcp_server
 - dns_server
 - log_client
 - log_server
 - nfs_client
 - nfs_server
 - repositories_server

See #205


actatux:
> Nice ! But where do you find all this time ?? 

In the evening. You see a single commit, but this was done on several days ;-)

Issue #220, opened 2020-05-14
dns_server multiple hosts with the same IP 


gauthierdelerce:
Hello,
When we generate the dns configuration, the same IP is defined for hostname and hostname-ice1-1 :
```
/var/named/forward:spartan198 IN A 10.10.1.150
/var/named/forward:spartan198-ice1-1 IN A 10.10.1.150
/var/named/reverse:150.1.10.10 IN PTR spartan198.bench.local.
/var/named/reverse:150.1.10.10 IN PTR spartan198-ice1-1.bench.local.
```
IMHO we should use a CNAME to define spartan198-ice1-1

Some tools as shine are puzzled with the IP defined used for the hostname also used for a different "A" record.

I'm testing this WA : 
```
sed -i "s/\(spartan.*\)-ice1-1 IN A.*/\1-ice1-1 IN CNAME \1/" forward
sed -i  "/spartan.*ice1-1/d" reverse
```

gauthierdelerce:
BTW, the new dns_server role requires python36-jmespath now.

actatux:
Hi @gauthierdelerce ,
The dependencies for Ansible are installed with the role core/ansible (will be renamed to core/bluebanquise soon).

actatux:
I'm thinking of a parameter `dns_server_zone_files = auto|static` for this role.
With `auto`, the zone files are created from templates.
With `static`, the zone files are copied from files, thus the admin can define the DNS zones with additional entries (CNAME, SRV, TXT, etc.). It would need a documentation of the mandatory entries (hostname, hostname-net).

oxedions:
This approach is understandable, other roles should also allow such things. I am think for example to Slurm role. Base role provides a minimum working configuration, but often local administrator is tuning slurm.conf file.
+1 on my side.

johnnykeats:
Same +1

gauthierdelerce:
Hi Bruno,
I'm not talking about the capability to add custom entries/parameters/forwarders/zones in the dns configuration ( which is always a good idea to ease customisation), but the fact that a bmc IP ( or the management IP on ice1 for nodes ) is defined twice:
we have 2 "A" records with the same IP

It's not clear if it's a issue, a "well known High End stack" is also defining 2 hostnames for the same IP, but 
- I don't think it's RFC compliant
- I don't know how to define nids/client in shine when two hostnames are returned by a nslookup

actatux:
Hi Gauthier,
[...]

PR #226, opened 2020-05-15
feat: add param dns_server_zone_files=auto|static

actatux:
Introduce new global parameter `dns_server_zone_files`. It allows to
define custom zone files `forward` and `reverse` when set to static.

 - auto: the zone files are created from templates
 - static: the zone files 'forward' and 'reverse' are copied from
   roles/core/dns_server/files/{forward,reverse}. You have to create the

The role default to auto when the parameter is not set or different than
static.

ci: add a new molecule scenario to test dns_server_zone_files=static
ci: move core/dns_server unit tests to dedicated actions
ci: start the dns service in all scenario of the role
ci: add Ubuntu 18.04 and 20.04 to the matrix

See #220

### DIFF
--- a/.github/workflows/molecule-actions-core-dns_server.yml
+++ b/.github/workflows/molecule-actions-core-dns_server.yml
@@ -1,0 +1,43 @@
+---
+name: core/dns_server unit testing
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'roles/core/dns_server/**'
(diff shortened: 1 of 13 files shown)


johnnykeats:
Nice idea.
Maybe too binary?
auto, static, or update.
update = generate reverse/forward from inventory, and allows user add custom reverse/forward added into reverse.custom and forward.custom files read by named.conf in addition to generated files.

actatux:
No more new feature before 1.5.

PR #230, opened 2020-05-16
Fix: log server seport deps

actatux:
Use of seport in the role requires:
 - libselinux-python
 - policycoreutils-python

PR for EL-like only (SELinux support).

Also removed an ineffective unit test.

### DIFF
--- a/roles/core/log_server/molecule/default/verify.yml
+++ b/roles/core/log_server/molecule/default/verify.yml
@@ -2,10 +2,6 @@
 - name: Verify
   hosts: all
 
-  vars:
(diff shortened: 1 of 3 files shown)


actatux:
While reviewing the unit tests, I noticed the role uses a template task for rsyslog.conf, while this is not a Jinja2 template (no variable).

I would propose a rework to use file tasks instead. I propose to drop the additional parameters in /etc/rsyslog.d/:
```
# Include all config files in /etc/rsyslog.d/
$IncludeConfig /etc/rsyslog.d/*.conf
```

PR #232, opened 2020-05-20
feat: merge equipment_profile with combine

actatux:
Because hash_behavior=merge will be deprecated with Ansible 2.10, move to the
combine filter to set the equipment_profile.

The default values are still defined in `equipment_profile`. To override any
default parameter for a specific equipment profile, define a new hash
`equipment_profile_override` with the parameter in the file
inventory/group_vars/equipment_*/equipment_profile.yml:

```
  equipment_profile_override:
    firewall: false
```

We need to call the combine filter explicitely, while hash_behavior was
transparent, thus the new role `common`.

### DIFF
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_HP_rusted/equipment_profile.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_HP_rusted/equipment_profile.yml
@@ -1,4 +1,4 @@
-equipment_profile:
+equipment_profile_override:
 
   access_control: true
   firewall: true
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_supermicro_management/equipment_profile.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_supermicro_management/equipment_profile.yml
@@ -1,4 +1,4 @@
-equipment_profile:
+equipment_profile_override:
 
   access_control: false
   firewall: false
--- a/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_supermicro_sandy_SMP/equipment_profile.yml
+++ b/resources/examples/multi_icebergs_cluster/inventory/group_vars/equipment_supermicro_sandy_SMP/equipment_profile.yml
@@ -1,4 +1,4 @@
-equipment_profile:
+equipment_profile_override:
 
   access_control: false
   firewall: false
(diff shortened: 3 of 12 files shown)


oxedions:
Dear @actatux 

No new features shall be added to core/advanced_core, as explained on slack.
We are discussing future of the stack.
Please do not spend time on new features, as they will probably not be accepted.

Ox

actatux:
This is still a draft. My understanding was no new feature for 1.3. We can discuss this one for 1.4.

johnnykeats:
> @actatux 
Discussed this internally, stack goes to "replace" HB, with no more modifications than merge of networks files into one, and simple rename of `all_equipments` to `equipment_global`.

Idea to use combine for eqp was discussed 2 weeks ago and discarded, brings too much users misunderstanding.

actatux:
> Discussed this internally, stack goes to "replace" HB, with no more modifications than merge of networks files into one, and simple rename of `all_equipments` to `equipment_global`.

For my knowledge, how will rename of `all_equipments` to `equipment_global` allow to replace `hash_behaviour=merge`?

johnnykeats:
Nothing directly related. Renaming will make users understand better the concept, as currently according to the feedback ox provided, users are confused.

Users will have to copy whole equipment_profile from global for each specific equipment group.

Simple change + lose a feature user often misunderstood -> no big lost

actatux:
> Users will have to copy whole equipment_profile from global to each specific equipment group.

That will be confusing and can lead to precedence issues.

How can we ensure Ansible will use the values from `equipment_foo/equipment_profile.yml` instead of `equipment_global/equipment_profile.yml`?

If the content of the new equipment_global should never be used by Ansible, we need to remove it from the inventory. It is still possible to provide this file in the documentation.

johnnykeats:
We are not sync:

/etc/bluebanquise/inventory/group_vars/all/equipment_global/
/etc/bluebanquise/inventory/group_vars/equipment_foo/

actatux:
OK, `all` comes before `equipment_foo`. It will work if users keep the inventory directory layout (and if they don't that's on them).
It's still confusing to get an inventory file that will never be used. We'll see how users react. Maybe add a header comment in the file to explain its purpose?

johnnykeats:
Understand your pov.
Global file is used as default profile, for homogeneous cluster or generic equipment that do not need a profile (basic item with ip but nothing else).

Agree on header, since users seems confused about precedence mechanism (lack of Ansible learning). Adding in the list.

actatux:
No more merge of equipment_profile.

oxedions:
Just a thought.

Since no more merge, it would be I think important that equipment profiles without `equipment_type == server` can handle nearly empty equipment_profile dict.

For a server, user copy the whole global file.
But for a switch, user could only use;

```
equipment_profile:
  equipment_type: switch

  equipment_authentication:
    user: ADMIN
    password: ADMIN
```

Since anything else is relevant. So code (pxe role, maybe others) needs to handle that.

PR #238, opened 2020-05-28
fix: open services in firewall's public zone

actatux:
If equipment_profile.firewall is true, the roles will add the relevant services to the firewall's default zone. In RHEL/CentOS, the firewall defaults to public zone.

For EL-like only. Can be extended to OpenSuse/Ubuntu later.
This PR focus on core and advanced-core roles only.

List of services per role:
 - dhcp_server/advanced_dhcp_server:
   - dhcp
 - dns_server:
   - dns
 - log_server:
   - syslog
   - rsh (for syslog 514/tcp)
 - nfs_server:
   - nfs
   - rpc-bind
 - pxe_stack:
   - http
   - tftp
 - repositories_server:
   - http
 - time:
   - ntp

I don't consider `ssh_master` because:

1. The ssh service is already in the services of the default zone
2. The role does not configure the sshd service

For each role, one can override the default zone (public) by setting `${rolename}_firewall_zone`. For example, if one wants to add the services of pxe_stack to the internal zone, he must set the variable `pxe_stack_firewall_zone: internal`. Configuration of the firewall zones must be managed by the user, either interactively or with a custom role. This is for possible future usage (addon role) and can be removed at any time, thus I let it undocumented. If an addon role should be added later, we will have to document this.

### DIFF
--- a/roles/advanced-core/advanced_dhcp_server/tasks/main.yml
+++ b/roles/advanced-core/advanced_dhcp_server/tasks/main.yml
@@ -19,6 +19,20 @@
         - "vars/{{ ansible_facts.os_family }}.yml"
       skip: true
 
+- name: "Add services to firewall's {{ advanced_dhcp_server_firewall_zone | default('public') }} zone"
+  firewalld:
+    zone: "{{ advanced_dhcp_server_firewall_zone | default('public') }}"
+    service: "{{ item }}"
+    immediate: "yes"
+    permanent: "yes"
+    state: enabled
+  when:
+    - ansible_facts.os_family == "RedHat"
+    - equipment_profile['firewall'] | default(false) | bool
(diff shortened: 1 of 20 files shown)


actatux:
I focused on RHEL/CentOS only to match the 1.3 window.
Tested with CentOS 8.

actatux:
> Just to understand, is it on purpose that firewall related tasks are before packages, or nothing related ?

Yes, the purpose was to open the ports as soon as possible. However, thinking about it, it may lead to issues for other roles if any rpm should drop custom services definition in /etc/firewalld/services/$myrpmservice.xml

actatux:
I can update the PR to move the firewalld tasks after the package tasks. Tell me.

actatux:
Discussed offline with @oxedions , no need to change right now.

PR #249, opened 2020-06-09
Oxedions/feat/scaling

oxedions:
New direct access method.

Replace existing but badly designed accelerated mode.

**Huge** performance improvement. Could even be better, but at a cost of greater complexity.
I personally think this is a good balance here (complexity vs performances).

Target is to cover all ranged roles:

* hosts_file
* dhcp_server
* dns_server
* ssh_master
* pxe_stack

hosts_file being the priority, as this file is often deployed on a massive number of hosts.

direct_access role **must** be executed before other roles to activate it in templates.
nodes_path variable must be set for the direct_access role to work. By default, nodes_path is /etc/bluebanquise/inventory/cluster/nodes/.

Direct access method only allows standard mg_group + eq_group yml structure, other structures cannot be handled right now.


oxedions:
PXE nodes file would benefit of a rework: instead of listing nodes one by one, nodes would be grouped by equipment profiles.

More logical, no more scaling issues on this file.

oxedions:
Stall. Waiting for another development.

roles/core/hosts_file/templates/hosts.j2
```diff
@@ -1,12 +1,11 @@
...
+{% import "macros/groups.j2" as macro_groups with context %}
 
-{# Macro to avoid redundancy #}
-{# This macro should cover all technical part, and be compatible with both normal and accelerated mode #}
-{% macro write_host_3(host,host_dict,host_node_main_network_interface,host_iceberg) %}
+{# -- Macro shared by both standard and direct_access methods. -- #}
+{% macro write_host(host,host_dict,host_node_main_network_interface,host_iceberg) %}
 
   {#- Write host #}
   {% if host_dict['network_interfaces'] is defined and not none and host_dict['network_interfaces'] and host_dict['network_interfaces'] is iterable %}
```
bouriquet, line 11: please remove "and not none" or replace this by either :

`host_dict['network_interfaces'] is defined and host_dict['network_interfaces'] is not none`

OR 

`(host_dict['network_interfaces'] | default(none)) is not none`

I think variables cannot be set to "none" or left empty (for initialization purpose ?) and should not be checked.

roles/core/hosts_file/templates/hosts.j2
```diff
@@ -59,55 +58,65 @@
...
+      {% endfor %}
+    {% endfor %}
+  {% endfor %}
+
 {% endif %}
-{{write_host_2(host,host_nic,host_node_main_network_interface,alias_list,host_bmc)}}
-{% endfor %}
-#}
+
 ## External hosts
```
bouriquet, line 116: This line leave an empty "## external hosts" line in /etc/hosts file when there is no external hosts, can you move this under the if statement, so this line is present only when external_hosts var is defined ?

roles/core/hosts_file/templates/hosts.j2
actatux:
```suggestion
  {% if host_dict['network_interfaces'] is defined and host_dict['network_interfaces'] is iterable %}
```
Should be enough?

PR #251, opened 2020-06-10
Add exclude parameter for RHEL and CentOS repositories in repository_client role

neilmunday:


neilmunday:
I missed off the `exclude` parameter for the OS repositories in RHEL/CentOS 8. This has now been fixed.

PR #256, opened 2020-06-14
Oxedions/feat/j2 removal

oxedions:
Current:

```
          network_interfaces:
            enp0s3:
              ip4: 10.10.0.1
              mac: 08:00:27:dc:f8:f5
              network: ice1-1
            ib0:
              ip4: 10.20.0.1
              network: interconnect-1
```

Will become:

```
          network_interfaces:
            - interface: enp0s3
              ip4: 10.10.0.1
              mac: 08:00:27:dc:f8:f5
              network: ice1-1
            - interface: ib0
              ip4: 10.20.0.1
              network: interconnect-1
```

With the following rules:

1. First item in the list is the hosts main resolution network (here, c001 for example will be on the same line than c001-ice1-1 in the hosts file).
2. First management network related item in the list will be the ansible mains ssh target interface (from ssh_master role), and also the main management network interface for the client (services_ip to target on client side).

If I wish direct resolution to be on interconnect-1 here, I simply move it first in the list. Ansible ssh will still use ice1-1 network, but ping c001 will be through interconnect-1.

This change provides:

* Greater powers to system administrator, as now, choice of main resolution **and** main management interface between each network is available.
* Removal of the need of global_network, i.e. simplification for hosts_file and dns_server.
* Simpler nic role, with better integration with Ansible nmcli module.
* Full removal of all j2_variables, massive code simplification.


johnnykeats:
Would go with the breaking change of network_interfaces first, and consider j2 removal later.

While network_interface change is logical, j2 removal seems to complexify code a lot. 

Like j2 variables the way they currently are. Simple. :blush:

oxedions:
Yeah, I have to agree with that. -_-

Spreading j2 variables everywhere makes code heavier, and if you need to update them, it makes things difficult. Let's keep j2 variables the way they are. I will only update the nic syntax.

actatux:
My only concern with the j2 variables is the ability to update those in user's inventory.

About the `network_interfaces` change, I believe this is a good move, technically. I'm more concerned by the introduction of a breaking change now, that will need extra work to mitigate the impact with existing end users. I think about custom roles that may depend on the current data structure. On the other hand, the sooner the better, so +1 with the change. ;-)

oxedions:
I was going to ask all of you your feedback on this, but since we already all agree :smile:

Ok, I will close this PR, and do all the changes in a new one.

I think impact on users should be light, as most custom roles rely mainly on j2_variables, and not directly on hosts network_interfaces, which is more used by the core. But not sure about that, just a guess.

Issue #259, opened 2020-06-24
FAILING TEST: core/time unit testing / ubuntu:18.04 - molecule test 


btravouillon:
The test is constantly failing for Ubuntu 18.04: https://github.com/bluebanquise/bluebanquise/runs/804781688?check_suite_focus=true

```
   TASK [time : package █ Install chrony tzdata packages] *************************
fatal: [client-ext-pool]: FAILED! => {"cache_update_time": 1590978383, "cache_updated": false, "changed": false,
"msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'chrony' 'tzdata''
failed: E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/d/dbus/libdbus-1-3_1.12.2-1ubuntu1.1_amd64.deb
  404  Not Found [IP: 91.189.88.142 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/d/dbus/dbus_1.12.2-1ubuntu1.1_amd64.deb
```

Indeed, the package does not exist anymore. It looks like Ubuntu maintainers remove some packages when they are superseded by a new release ([which occurred last week for dbus in 18.04](https://launchpad.net/ubuntu/+source/dbus/1.12.2-1ubuntu1.1/+publishinghistory))

My first thought was to update the apt cache of the docker instance in the prepare step of the molecule test. However, a user might hit the same issue in the field since the stack run the package tasks with `update_cache: no` (default).

Any thought?

oxedions:
I think that for now, we should disable Ubuntu related molecule tests.
The stack do not officially support Ubuntu, and there was no request to support it, even if many parts are ready.

If one day, some team or company ask for Ubuntu, we will reintroduce them.

actatux:
OK. I will disable this test for 18.04 then.
I propose to keep other tests while they succeed: it would prevent any regression since Ubuntu might be in the scope later.

actatux:
Similar issue **test core/conman unit testing / ubuntu:20.04 - molecule test**: https://github.com/bluebanquise/bluebanquise/runs/932326901?check_suite_focus=true

I will implement @oxedions proposal and disable all the molecule tests for Ubuntu since it is not supported.

PR #261, opened 2020-06-26
ci: remove Ubuntu 18.04 from molecule core/time

actatux:
Fixes: #259


oxedions:
Since JK is off, merging.

PR #262, opened 2020-06-26
Fix disklessset

bouriquet:
- disable cache
- patch nfs staging to golden process
- removing hard coded releasever in dnf install commands
- disable selinux in livenet kernel cmdline

Known issues :
- image creation fail when repositories_releasever var is not set (dnf var should be generated with distribution_major_version in repositories_client role instead of removing the file)


oxedions:
For now, this makes the role works.

However, we will need to plan a clean of the script, and a way to get back the nfs cache mechanism, as it improves a lot images sizes.

PR #252, opened 2020-06-11
Oxedions/fix/cleanresources

oxedions:
Clean of resources.

Also includes fallback to replace hash behavior, as mainly related to inventory structure.

PR #263, opened 2020-06-28
fix: use --limit in bluebanquise-playbook

actatux:
When a list of nodes is provided with -n, run ansible-playbook with the
parameter --limit.

Remove support of --extra-vars target=nodeset. If
`hosts: {{ target }}` is found in the playbook, print a deprecation
error and exit.

Related to changes in #252.

### DIFF
--- a/tools/bluebanquise-playbook
+++ b/tools/bluebanquise-playbook
@@ -50,9 +50,9 @@ fi
 
 playbook_yml="${ANSIBLE_CONFIG}/playbooks/${playbook}.yml"
 if [[ $(grep -c 'hosts:.*{{.*target.*}}' ${playbook_yml}) -ne 0 ]]; then
-    echo -e "\e[31mWARNING: Usage of {{ target }} in hosts is deprecated."\
+    echo -e "\e[31mERROR: Usage of {{ target }} in hosts is deprecated."\
          "Please update your playbook.\e[0m"
-    sleep 3
+    exit 1
 fi
 
 ansbook_cmd=$(which ansible-playbook)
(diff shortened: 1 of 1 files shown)


tools/bluebanquise-playbook
```diff
@@ -48,6 +48,13 @@ elif [[ ! -f "${ANSIBLE_CONFIG}/playbooks/${playbook}.yml" ]]; then
     exit 1
 fi
 
+playbook_yml="${ANSIBLE_CONFIG}/playbooks/${playbook}.yml"
+if [[ $(grep -c 'hosts:.*{{.*target.*}}' ${playbook_yml}) -ne 0 ]]; then
+    echo -e "\e[31mWARNING: Usage of {{ target }} in hosts is deprecated."\
+         "Please update your playbook.\e[0m"
+    sleep 3
```
actatux, line 55: Instead of deprecation, we can also drop the support of target:
```suggestion
    echo -e "\e[31mERROR: Usage of {{ target }} in hosts is deprecated."\
         "Please update your playbook.\e[0m"
    exit 1
```

bouriquet: your script also override "-l" option which is the short option for --limit, probably not a good idea to override this

actatux: Indeed, I will replace this with --list. Thanks.

actatux: The script is fairly simple and use bash getopts, thus I cannot easily switch to `--list`. The --limit is already handled with `-n <nodeset>`, there is no need to use ansible-playbook's `-l` with the command.

oxedions: -n comply with bootset command, so ok for me.

oxedions: I would raise an error with target, since it is more than deprecated: it proves to be dangerous with playbooks.

PR #264, opened 2020-06-29
Added support for multiple exporters

sobrase:
Automatic generation of prometheus.yml for the exporters, and installation of the exporters


roles/addons/prometheus_client/tasks/main.yml
```diff
@@ -1,54 +1,33 @@
----
+#- debug:
```
oxedions, line 1: Can you remove these comments ?

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -6,8 +6,31 @@
     state: present
   with_items:
     - prometheus
+    - grafana
     - alertmanager
     - ipmi_exporter
+    - freeipmi
+    - snmp_exporter
+
```
oxedions, line 14: Can you also add Karma ?

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -6,8 +6,31 @@
...
+    - net-snmp
+    - net-snmp-utils
+    - net-snmp-libs
+    - net-snmp-devel
+  when:
+    - monitoring.exporters.snmp_exporter.with_generator is defined      
+    - monitoring.exporters.snmp_exporter.with_generator
+
+
+
```
oxedions, line 32: Can you clean multiple empty lines spaces like these ?

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -78,12 +85,28 @@
...
+    owner: ipmi_exporter
+    group: ipmi_exporter
     mode: 0640
   tags:
     - template
 
+#- name: Deploy Slurm Exporter Binary files
+#  copy:
+#    src: prometheus-slurm-exporter
+#    dest: /usr/bin/prometheus-slurm-exporter
```
oxedions, line 97: Can you remove these comments ?

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -92,6 +115,8 @@
     - prometheus
     - alertmanager
     - ipmi_exporter
+    - snmp_exporter
+#    - prometheus-slurm-exporter
```
oxedions, line 119: same

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -103,6 +128,8 @@
     - prometheus
     - alertmanager
     - ipmi_exporter
+    - snmp_exporter    
+#    - prometheus-slurm-exporter
```
oxedions, line 132: same

roles/addons/prometheus_server/tasks/main.yml
actatux: What extra repo do we need to install karma?

actatux:
It's a bit off-topic, but I can't find any documentation for monitoring.exporters in this role. There is an example in prometheus_server, but there is no monitoring.exporters.*.package.
Could we use this PR to enhance the documentation of this role?

roles/addons/prometheus_server/tasks/main.yml
oxedions: Karma is autonomous. We will just need to add it as a package like prometheus and the others, in the bluebanquise infrastructure. We only package official binary, and add a service file with it.
This is not very conventional, but these tools are not that easy to build, and since these static binaries are provided by very big teams, I think we can use binaries provided on their github, to also ease support in case of issue.

oxedions:
> It's a bit off-topic, but I can't find any documentation for monitoring.exporters in this role. There is an example in prometheus_server, but there is no monitoring.exporters.*.package.
> Could we use this PR to enhance the documentation of this role?

We are on it 😊

roles/addons/prometheus_server/tasks/main.yml
actatux: Should we open an issue in bluebanquise-infrastructure for tracking?

oxedions: Good idea. Let me do it.

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -6,8 +6,31 @@
...
     - ipmi_exporter
+    - freeipmi
+    - snmp_exporter
+
+#pour generateur de oid walk pour snmp_exporter     
+- name: Package
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
```
osmocl, line 21: Please use "loop" 
https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-items

roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -6,8 +6,31 @@
     state: present
   with_items:
```
osmocl, line 7: Idem.
https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-items

roles/addons/prometheus_server/tasks/main.yml
bouriquet: Do not use loop in package related modules, Ansible recommend passing the list directly in the name

From Ansible documentation https://docs.ansible.com/ansible/latest/modules/yum_module.html :
"When used with a loop: each package will be processed individually, it is much more efficient to pass the list directly to the name option."

PR #265, opened 2020-06-30
Add ignore-client-uids parameter in dhcpd.conf

bouriquet:


roles/advanced-core/advanced_dhcp_server/templates/dhcpd.conf.j2
```diff
@@ -17,6 +17,7 @@
 
  # For option 82
  stash-agent-options true;
+ ignore-client-uids on;
```
oxedions, line 20: This parameters violates the specification. I would prefer to have it integrated in an if, like this:

```
{% if dhcp_settings.ignore-client-uids is defined and dhcp_settings.ignore-client-uids == true %}
  ignore-client-uids on;
{% endif %}
```

And set at false by default in the example inventories.

actatux: or `ignore-client-uids {{ dhcp_settings.ignore-client-uids | default('off') }}` ?

PR #269, opened 2020-06-30
Feat: Split inventory to isolate j2_variables

oxedions:
Split inventory into inventory folder, as default inventory
and internal folder, that contains j2_variables and internal
variables of the stack.
Also moved security.yml in general_settings.


internal/group_vars/all/internal_variables.yml
```diff
@@ -4,6 +4,4 @@ iceberg_naming: iceberg
 equipment_naming: equipment
 management_networks_naming: ice
 master_groups_naming: mg
-
 managements_group_name: mg_managements
```
actatux, line 7: I believe we should keep this file in the user's inventory to keep the configuration in a single place.

resources/examples/multi_icebergs_cluster/inventory/group_vars/all/j2_variables/internal_variables.yml
```diff
@@ -1,9 +0,0 @@
-### NAMING
-
-iceberg_naming: iceberg
-equipment_naming: equipment
-management_networks_naming: ice
-master_groups_naming: mg
-
-managements_group_name: mg_managements
-
```
actatux, line 9: I believe we should keep this file in the user's inventory to keep the configuration in a single place.

oxedions:
I checked with the other users, and ok for not moving internal_variables. Updating, and merging.

PR #272, opened 2020-07-01
fix: make converter compatible with older PyYAML

actatux:
CentOS 8 ship with PyYAML 3.12 while I used some features introduced
with 5.1, which are incompatible. See yaml/pyyaml#265.

It appears I tested it in my virtualenv setup with latest PyYAML 5.3.1.... :-1:

### DIFF
--- a/tools/inventory-converter-1.3-network_interfaces.py
+++ b/tools/inventory-converter-1.3-network_interfaces.py
@@ -60,13 +60,13 @@ def main():
 
     with open(hostsfile, 'r') as fd:
         try:
-            inventory = yaml.load(fd, Loader=yaml.FullLoader)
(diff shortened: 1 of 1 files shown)


actatux:
Without this patch, the code fails on CentOS8 while converting the inventory files:

```
[root@management1 bluebanquise]# ./tools/inventory-converter-1.3-network_interfaces.py inventory/cluster/nodes/computes.yml 
Traceback (most recent call last):
  File "./tools/inventory-converter-1.3-network_interfaces.py", line 77, in <module>
    main()
  File "./tools/inventory-converter-1.3-network_interfaces.py", line 63, in main
    inventory = yaml.load(fd, Loader=yaml.FullLoader)
AttributeError: module 'yaml' has no attribute 'FullLoader'
```

Issue #182, opened 2020-04-16
General features requests dashboard


oxedions:
# General features planed for 1.4.0

## Focus on

**Core**
- [ ] bootset dedicated kernel params (ability to add dedicated kernel params at bootset invocation. Nodes files already have this variable).
- [ ] nmcli replacement for nic role (replace current templates by nmcli module).
- [ ] Advanced routing for network (default gateway in case of multiple, routing in multiple icebergs, etc).

**Addons**
- [ ] Rewrite diskless role to handle plugins.
- [ ] Ssh jump in multi iceberg managements (allows a single point of management or playbook execution).

**Tools**
- [ ] Enhance bluebanquise-playbook with spread task capability (ability to use command line or YAML file to spread a playbook execution over executors, in case of massive deployment).

**Doc**
- [ ] Enhance Advanced DNS server role readme.
- [ ] bootset man.
- [ ] Add more stories.
- [ ] Improve report role.

## If possible
**Core**
- [ ] Redo PXE part for Fedora, OpenSuse, Debian and Ubuntu. Only support PXE of these distribution over client nodes for now.

**Addons**
- [ ] Clean clone addon.
- [x] Basic Lmod role.

# General features planed for 1.3

1.3 is targeted to be a quality release. No major changes in core.

**Architecture/General**
- [x] Switch to replace HB
  - [x] merge networks (#252)
  - [x] rename all_equipment (#252)
  - [x] header in global file
  - [x] handle empty dict if not server
- [x] Add j2 into github root inventory (allows updates) (#269)
- [x] Tasks naming convention

**Core**
- Repositories
  - [x] Advanced core repositories
  - [x] Deprecate external_repositories (#270)
- [x] Role to enforce equipment_profile
- bootset
  - [x] status
- DNS
  - [x] Advanced DNS
- [x] Firewall handling in roles (#238)
- [x] PXE nodefile re-ordering

**Addons**
- [x] Finish and validate Prometheus roles (monitoring, @oxedions )
- [x] Login nodes addon role
- [x] Power management tool

**Doc**
- [x] password generation
- [x] facts/include/etc missing from training
- [x] Clean example inventories (#252)
- [x] equipment_profiles values are not properly explained in the documentation, and available values are not provided (@oxedions )

# General features requested delayed

- [ ] [DHCP failover](https://kb.isc.org/docs/aa-01356)
- [ ] Rewrite ldap_client/server role (and rename it) to use 389 and handle cockpit related plugin
[...]

PR #277, opened 2020-07-10
Oxedions/fix/core simplification

oxedions:
Simplify core.

Repositories and DNS server moved to advanced, and should be able to receive more features in future.
New basic roles for core (without many features, be careful when using).


oxedions:
Seems good. I will squash once validated before merging.

oxedions:
Advanced roles may need a little more documentation in the readme files. Can be done later.

actatux:
Hello,

Before reviewing, why did you change your mind about these 2 roles? We agreed to let these in core in [this thread](https://github.com/bluebanquise/bluebanquise/issues/182#issuecomment-640064232).

oxedions:
I remember :-)

There are 2 reasons:

1 - I had some issues with the advanced roles: one was not compatible with the diskless addon, the second one was not compatible with the HA that my company is implementing. Both issues could simply be bypassed for an advanced system administrator, but I want core to work "as is", while I understand advanced core need advanced knowledge.
2 - Some users reported still too much complexity on these roles, which was not compatible with the didactic goal of the stack.

So I choose to move them into advanced, and create very basic ones for the very basic usage. The core ones will keep very basic or naive, while advanced ones will be able to offer more and more features as needed in the future and will concentrate developments focus. 

Same happen for dhcp server before.

actatux:
1 - I would be pleased to have further details since we are working for the same company and I'm not aware of these issues.

2 - I will review then, but I would tend for very basic and simple roles. The current proposal is too complicated and, at first read, too close to the advanced roles.

roles/core/dns_server/tasks/main.yml
```diff
@@ -42,61 +42,93 @@
...
+    group: root
+    mode: 0644
+  register: zoneconf
   tags:
-    - internal
+    - template
 
-# If dns_ip is a list, pick the first value as master DNS
-- name: set_fact ░ Set master DNS server - Pick first element of dns_ip
+- name: set_fact ░ Set SOA Serial for forward zone
```
actatux, line 81: redundant with previous task line 62.

roles/core/dns_server/templates/named.conf.j2
```diff
@@ -74,43 +64,17 @@ zone "." IN {
   file "named.ca";
 };
 
-{# Is the host the master DNS? #}
-{% if dns_master in inventory_all_ipv4_addresses %}
```
actatux, line 78: Beware, users might end with several master dns if they run the role on two management nodes. You need to ensure there is a single dns server for all the cluster.

roles/core/dns_server/vars/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-dns_server_role_version: 1.1.1
+dns_server_role_version: 1.1.2
```
actatux, line 2: This is a new role, should be 1.0.0.

johnnykeats:
Hello Ox.
As promised, will have a look. ;)

actatux:
> 1 - I had some issues with the advanced roles: one was not compatible with the diskless addon,

I found out this is repositories_client. I opened #283 which should fix the issue in the role diskless.

johnnykeats:
@oxedions 

Ok, got it now.
You are again trying to adapt the stack to your company needs. You accepted many things you were against of because of external pressures, which makes us unhappy.
1.2 version was stable, now it is a mess.

BlueBanquise is an **independent** stack. We had excellent feedback with the original philosophy and architecture.

Please revert and remove all the things you do not agree with before going further, then only will I review.

oxedions:
So be it. I was as always way too kind.

I should have done this long time ago.

May take few days.

PR #231, opened 2020-05-18
WIP feat: firewall rhel

actatux:
Work in progress. Draft for review/advice.

Add new role core/firewall. Only support RHEL/CentOS.

Also add services to firewall's internal zone:
 - dhcp
 - dns
 - syslog
 - rsh (for syslog 514/tcp)
 - nfs
 - rpc-bind
 - http
 - tftp
 - ntp

PR #282, opened 2020-07-13
Oxedions/feat/scaling new nic

oxedions:
New direct access method.

Replace existing but badly designed accelerated mode.

Huge performance improvement. Could even be better, but at a cost of greater complexity.
I personally think this is a good balance here (complexity vs performances).

Target is to cover all ranged roles:

    hosts_file
    dhcp_server
    dns_server
    ssh_master
    pxe_stack

hosts_file being the priority, as this file is often deployed on a massive number of hosts.

direct_access role must be executed before other roles to activate it in templates.
nodes_path variable must be set for the direct_access role to work. By default, nodes_path is /etc/bluebanquise/inventory/cluster/nodes/.

Direct access method only allows standard mg_group + eq_group yml structure, other structures cannot be handled right now.

### DIFF
--- a/roles/advanced-core/direct_access/defaults/main.yml
+++ b/roles/advanced-core/direct_access/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+nodes_path: "/etc/bluebanquise/inventory/cluster/nodes"
--- a/roles/advanced-core/direct_access/tasks/main.yml
+++ b/roles/advanced-core/direct_access/tasks/main.yml
@@ -1,0 +1,11 @@
+---
+
+- name: "include vars █ Include vars from {{ nodes_path }}"
+  include_vars:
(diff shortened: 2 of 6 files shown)


actatux:
I'm a bit lost. Less than two months ago, [it was against the design of the stack to share variables between two roles](https://github.com/bluebanquise/bluebanquise/pull/231#issuecomment-633898623).

johnnykeats:
Would suggest to create an Advanced hosts file role, with autonomous files read.

This feature should not be in the core as code is difficult to read for new users.
If other roles need same feature, only Advanced ones should benefit from it.

oxedions:
This PR is on standby, and wait for the 1.3.0 GA, so to be included as a fix in 1.3.1.

oxedions:
Not really needed anymore since new versions is much faster. Will keep the idea in mind if needed later.

PR #283, opened 2020-07-14
fix: define dnf variables in disklessset

actatux:
It is possible to define dnf variables $environment and $releasever in
the repositories_client role, to set the baseurl of a repository. When
disklessset run dnf with a new installroot, dnf is not able to get the
values of those variables and eventually fails.

Since the script relies on the host configuration to get repos
information, make the script a jinja2 template file and define the
$environment and $releasever in the installroot from the host inventory.

If repositories_releasever is not defined, resolve to the releasever of
the host.

### DIFF
--- a/roles/addons/diskless/tasks/main.yml
+++ b/roles/addons/diskless/tasks/main.yml
@@ -20,8 +20,8 @@
       skip: true
 
 - name: disklessset tool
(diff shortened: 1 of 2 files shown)


actatux:
dnf variables are removed with #285.

PR #289, opened 2020-07-22
New SSSD role where several servers can be configured (SSL or not)

gavillom:


roles/addons/sssd/readme.rst
```diff
@@ -0,0 +1,46 @@
...
+            - ldap://my-ldap-server-2
+          time_ip: 10.10.0.1
+          log_ip: 10.10.0.1
+
+The following example configure LDAP secured server *my-ldap-server-1* as first server and *my-ldap-server-2* as second server.
+
+Changelog
+^^^^^^^^^
+
+* 1.3.0: Support multiple and secured connections to LDAP servers. <mathis.gavillon@atos.net>
```
oxedions, line 43: Can you increment to 1.0.2 ?

roles/addons/sssd/templates/sssd.conf.j2
```diff
@@ -0,0 +1,30 @@
...
+id_provider = ldap
+auth_provider = ldap
+#access_provider = ldap
+access_provider = permit
+chpass_provider = ldap
+ldap_uri = {{networks[j2_node_main_network]['services_ip']['authentication_servers']|join(',')}}
+ldap_search_base = {% for dc in domain_name.split('.') %}{% if not loop.first %},{% endif %}dc={{dc}}{% endfor %}
+
+ldap_tls_reqcert = never
+ldap_auth_disable_tls_never_use_in_production = true
```
oxedions, line 13: To enable tls, I think these 2 parameters skip it. Would worth investigating 😊

roles/addons/sssd/readme.rst
gavillom: Done

roles/addons/sssd/templates/sssd.conf.j2
gavillom: Done

oxedions:
Thanks a lot. Let use this role as base also for other authentications API. 😊

PR #299, opened 2020-08-03
[Diskless] Missing kernel RPM in livenet image

hmeAtos:
Hello,
When you want to install kernel modules in the diskless image: Luster, MOFED or other. 

We have this error message:
An exception occurred during task execution. To see the full traceback, use -vvv. 
The error was: FileNotFoundError: [Errno 2] No such file or directory: '/lib/modules/4.18.0-193.6.3.el8_2.x86_64/modules.builtin'
fatal: [pm5-nod39]: FAILED! => changed=false
msg: '[Errno 2] No such file or directory: ''/lib/modules/4.18.0-193.6.3.el8_2.x86_64/modules.builtin'''

To fix this issue we must add RPM Kernel when creating the livenet image.

Regards,
@hmeAtos

PR #301, opened 2020-08-05
Enhance disklessset livenet

btravouillon:
This PR implements some fixes and quick and dirty enhancements for the disklessset tool.

I fixed some bugs identified during testing of the tool:
1. 'Remove an image' action was not implemented
2. livenet image size was not user-friendly (ask for 1MB to create a 1GB image)
3. crash in load_kernel_list() when more than one file was found in the directory
4. over-consumption of disk space with the image_working_directory being copied next to the squashfs.img
5. crash in menus or submenus when no kernel or image are found
6. install disklessset after its requirements: a user reported crash of disklessset because these directory don't exist (might be HA related)

I added the features below:
1. inject a custom SSH public key to the livenet image: temporary solution to provide SSH key-based authentication, which allows to boot diskless nodes then run ansible-playbook at scale. Long term solution would be to inject the values of authentication.ssh_keys in the image, or use cloud-init.
2. add image information to /etc/os-release: useful to know which image the node loaded at boot.
3. select additional packages in diskless image: kind of dynamic profile based on user selection, trying to avoid future PR/issue requesting additional packages in one of the 3 static profiles. (see #299).

### DIFF
--- a/roles/addons/diskless/files/disklessset.py
+++ b/roles/addons/diskless/files/disklessset.py
@@ -11,13 +11,15 @@
 # 2019 - Benoît Leveugle <benoit.leveugle@sphenisc.com>
 # https://github.com/bluebanquise/bluebanquise - MIT license
 
-# Import dependances
-from ClusterShell.NodeSet import NodeSet
-from argparse import ArgumentParser
-import yaml
+# Import dependencies
 import os
 import crypt
+import shutil
+from argparse import ArgumentParser
 from datetime import datetime
+
+import yaml
+from ClusterShell.NodeSet import NodeSet
 
 
 # Colors, from https://stackoverflow.com/questions/287871/how-to-print-colored-text-in-terminal-in-python
@@ -181,8 +183,7 @@ print('BlueBanquise Diskless manager')
 print(' 1 - List available kernels')
 print(' 2 - Generate a new initramfs')
 print(' 3 - Generate a new diskless image')
-print(' 4 - Manage/list existing diskless images')
-print(' 5 - Remove a diskless image')
+print(' 4 - Manage existing diskless images')
 
 main_action = str(input('-->: ').lower().strip())
 
@@ -358,6 +359,7 @@ elif main_action == '4':
     print(' 2 - Manage kernel of an image')
     print(' 3 - Create a golden from a staging NFS image')
     print(' 4 - Manage hosts of an NFS image')
+    print(' 5 - Remove an image')
     sub_main_action = str(input('-->: ').lower().strip())
 
     if sub_main_action == '1':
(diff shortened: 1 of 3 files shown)


oxedions:
Dear @btravouillon 

Thanks a lot for this PR.

I am ok with 1, 2 and 3. Not with 4, because I let the working dir on purpose. Let me explain the idea:

Step one, generate a basic image (bootable + ssh).

* User create an image, with "small" profile, set size according to needs to work
* Image is created, and squashfs is generated, and copied into html folders, with its working dir (raw image)

Step two, customize image:

* User mount raw image from diskless tool
* User can apply playbooks inside the raw image using:
 ansible-playbook -i /etc/bluebanquise/inventory/ -i /etc/bluebanquise/internal/ -i /etc/bluebanquise/diskless/ diskless.yml -e "j2_current_iceberg=iceberg1 j2_node_main_network=ice1-1 start_services=false" --skip-tags specialize
* User can tune other parts of the image, according to needs
* User use diskless tool to repack image, which generate a new squashfs

Step tree, optional, reduce image size:

* User use diskless tool to reduce image size.
* Option 1: user set free disk that should be left, and tool reduce image size to "/ root size + free asked space"
* Option 2: more aggressive, but lock image: user ask for removal of dnf in the Option1 process, to possibly gain around 100mb. Not sure about that part, need investigations.

This is why I would like to keep the raw image 😊

oxedions:
Would it be possible to submit the PR without option 4 ?

btravouillon:
Understood. I would have unsquashfs the image to a new working dir, make the change, then replace the squashfs. Is it ok to replace original copy with a move? The purpose is to clean `image_working_directory`.

hmeAtos:
_> 1. 'Remove an image' action was not implemented_
--> ok for me, thanks for the implementation

_> 2. livenet image size was not user-friendly (ask for 1MB to create a 1GB image)_
--> Good for me, it will cause less error

_> 3. crash in load_kernel_list() when more than one file was found in the directory_
-->ok for me, thanks for the implementation

_> 4. over-consumption of disk space with the image_working_directory being copied next to the squashfs.img_
--> not sure it's a good idea, we need this 'reference' directory for the customisation 

_> 5. crash in menus or submenus when no kernel or image are found_
--> ok for me, it was a issue

_> 6. install disklessset after its requirements: a user reported crash of disklessset because these directory don't exist (might be HA related)_
I agree

oxedions:
Dear @hmeAtos 

Point 4 was discussed internally, we can unsquashfs so I think it is ok to remove the raw image :-)

hmeAtos:
> Dear @hmeAtos
> 
> Point 4 was discussed internally, we can unsquashfs so I think it is ok to remove the raw image :-)

ok perfect !

PR #306, opened 2020-08-12
Call imgstat in ipxe files to get image report

oxedions:
Add imgstat to see useful data before booting.

See https://ipxe.org/cmd/imgstat

I will update the other files if others are ok with this small add.


oxedions:
In fact, I had issues with a server that was downloading files, but network was somehow broken, and final files on ipxe side were truncated. Using imgstat I later seen that the size before booting on client side was far smaller than the ones on server side...

oxedions:
Done. Added into other files and diskless.

Ready for review :-)

PR #312, opened 2020-08-15
fix: remove equipment_profile key from inventory

btravouillon:
Note: this is a proposal to open the discussion before 1.3 GA.

While porting my test bed from 1.2 to 1.3-rc1, I had to modify my inventory files because of the major changes we introduced (mainly due to removal of `hash_behaviour = merge` due to its deprecation in Ansible). I was copying the 
values of `group_vars/all/equipment_all/equipment_profile.yml` to the relevant `group_vars/equipment_type*/equipment_profile.yml` when I asked myself what is the purpose of the `equipment_profile` key now.

The `equipment_profile` key defined in equipment_profile.yml files was useful when the stack was using `hash_behaviour = merge`. Now, it prevents to inherit values from the global configuration file `groups_vars/all/equipment_all/equipment_profile.yml`, with no obvious added value (at least to me, I might miss it :sweat_smile:).

This change removes this key from inventory examples and usage of this key in the roles.
It does not change the logic of equipment_profile files.

With the breaking changes we are introducing in 1.3, I guess it is the right time to think about this. Converting an equipment_profile.yml file from a 1.2 inventory to a 1.3 inventory could be as easy as: remove `equipment_profile` key and decrease indentation by one level.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,15 +141,15 @@ can be defined in the role vars. Example: packages names, services names, paths,
 in dedicated files or if needed in a new one. When possible, variables should be optional.
 In any cases, they must be documented in the role readme (and if needed provided commented in the example inventories).
 
-4. All variables related to an equipment_profile should go in group_vars/all/all_equipment/ (global) or in group_vars/equipment_X/ with X the equipment profile name (dedicated).
+4. All variables related to an equipment_profile should go in group_vars/all/all_equipment/ (global) or in group_vars/equipment_X/ with X the equipment profile name (dedicated). These variables must be prefixed by **ep_**.
 
 5. All variables containing jinja2 code must be prefix with *j2_* and stored in group_vars/all/j2_variables/.
 j2_ variables are core of the stack, and should be manipulated with care.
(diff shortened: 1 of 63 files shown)


oxedions:
Hello 😊 

Took me time to compute.

I agree with this strategy, as we lost a lot with merge deprecation. It seems now we should avoid big dictionaries.

To keep the equipment_profile philosophy, i.e. variables dedicated to equipment_profile groups, and that should never be used outside of it, I would prefix all these values with 'ep_' .

Also, I would take this opportunity to do the same for authentication, simply using 'authentication_' prefix.

Would that be ok for all ?

btravouillon:
Command to update inventory files from 1.2 to 1.3:
```
$ sed -i -e 's/^equipment_profile:$/---/' \
             -e 's/^  //' \
             -e 's/^\([a-z]\)/ep_\1/' \
  inventory/group_vars/all/equipment_all/equipment_profile.yml \
  inventory/group_vars/equipment_type*/equipment_profile.yml
```

oxedions (approved):
Dear @btravouillon 

Seems all good to me. I will merge, then redeploy the whole stack to check everything.

Ox

PR #324, opened 2020-08-20
Fix: Update prometheus_server readme.rst

sobrase:
Added yml file for prometheus_server role used to configure prometheus

### DIFF
--- a/roles/addons/prometheus_server/readme.rst
+++ b/roles/addons/prometheus_server/readme.rst
@@ -15,15 +15,20 @@ Alertmanager is available at http://localhost:9093
 
 Node_exporter is available at http://localhost:9100
 
-Put the following under your inventory/group_vars/all/addons/monitoring.yml ::
+In order to allow the role to deploy, you need to add the following 
+content under a file inventory/group_vars/all/addons/monitoring.yml :
 
-prometheus:
-  scrape_interval: 1m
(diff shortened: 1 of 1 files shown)


oxedions:
Can you add a yaml code block for this ?

Also please remove the  :: and only add a single :

roles/addons/prometheus_server/readme.rst
```diff
@@ -15,6 +15,21 @@ Alertmanager is available at http://localhost:9093
 
 Node_exporter is available at http://localhost:9100
 
+In order to allow the role to deploy, you need to add the following 
+content under a file inventory/group_vars/all/addons/monitoring.yml :
```
btravouillon, line 19: I would rather prefer to set defaults/main.yml, which would allow the role to deploy out-of-the-box.

osmocl:
Hello
Sorry to bother you :)
I was wondering if you could provide some exporters examples in the readme too.

```
# cat monitoring.yml
monitoring:

  exporters:
    node_exporter:
      package: node_exporter
      service: node_exporter
[...]

roles/addons/prometheus_server/readme.rst
oxedions: Deal !

oxedions:
Done in 330.

PR #326, opened 2020-08-20
doc: prometheus_client cleanup

sobrase:
Deleted unwanted example file for prometheus_client and added it to the readme file.


roles/addons/prometheus_client/readme.rst
```diff
@@ -11,6 +11,33 @@ Instructions
 
 Node_exporter is available at http://localhost:9100
 
+exporters configuration file to add to inventory/group_vars/equipment_type_X
+depending on where the exporter should be:
```
btravouillon, line 15: Does the role work out-of-the-box without these parameters?

oxedions: Without this, the role will do nothing.

btravouillon: Can we consider adding this to defaults/main.yml?

oxedions: No, impossible for these ones. These need to be manually tuned for each kind of equipment, no defaults possible here.

oxedions: Ho ok got it, you mean an empty configuration. Will do it.

oxedions:
Done in 330.

Issue #328, opened 2020-08-22
pxe_stack - ScannerError: while scanning a plain scalar


btravouillon:
## Description

Error with the YAML parser during non reg tests with EL7. It looks like the new structure of network_interfaces triggers a parsing error with MAC addresses (because of columns in the string).

```
TASK [pxe_stack : template █ Generate /etc/bluebanquise/pxe/nodes_parameters.yml] *****************
Saturday 22 August 2020  03:10:54 +0000 (0:00:02.889)       0:00:34.188 ******* 
fatal: [management1]: FAILED! => changed=false 
  msg: |-
    ScannerError: while scanning a plain scalar
      in "<unicode string>", line 1, column 33:
        {u'interface': u'ens5', u'mac': u'08:00:27:0d:44:90', u'network' ...
                                        ^
    found unexpected ':'
      in "<unicode string>", line 1, column 37:
        {u'interface': u'ens5', u'mac': u'08:00:27:0d:44:90', u'network': u' ...
                                            ^
    Please check http://pyyaml.org/wiki/YAMLColonInFlowContext for details.
```
## Inventory
```
      hosts:
        c001:
          network_interfaces:
            - interface: ens5
              ip4: 10.10.3.1
              mac: 08:00:27:0d:44:90
              network: ice1-1
```
Note the inventory is defined with block style but Ansible converts it to flow style.

## Info

BlueBanquise 1.3-RC2
CentOS Linux release 7.8.2003 (Core)
ansible 2.9.10
python 2.7.5
PyYAML-3.10

PR #331, opened 2020-08-22
fix: drop macro node_main_nic from pxe_stack

btravouillon:
The macro node_main_nic returns a string encoded as unicode, which
breaks the rendering with Python 2. Since the macro is used once in the
template, remove it and use its logic in the for-loop.

Fixes: #328

Let's use Jinja2 for what it does best: templating. :relieved:


oxedions (approved):
Dear @btravouillon 

Thanks a lot for this PR 😊 

This is indeed a remaining of the previous inventory network_interfaces structure, and is not needed anymore.

Merging.

PR #343, opened 2020-08-26
Fix: Update inventory examples

oxedions:
Take into account possibility to only override desired values.

Note: I sometime left some values that are also the same than in global, but most of the time, I consider better to "force" security related values.


osmocl (approved):
Ok for me :)

oxedions:
Yes, because if the monitoring addon is not used, this file is not needed.

Since monitoring is not in the core, but in addons, I prefer to remove it 🙂

Issue #329, opened 2020-08-22
time - AnsibleError: template error while templating string: no test named 'gt'.


btravouillon:
## Description

Non reg testing with EL7. The comparison `gt` (greater than) does not exist:
```
TASK [time : template █ Generate /etc/chrony.conf] *************************************************
Saturday 22 August 2020  03:39:47 +0000 (0:00:00.729)       0:00:02.370 ******* 
fatal: [management1]: FAILED! => changed=false 
  msg: |-
    AnsibleError: template error while templating string: no test named 'gt'. String: 
[...]
    # Serve time even if not synchronized to a time source.
    # Fairly unreliable time source
      {% if iceberg_level is defined and iceberg_level is not none %}
        {% if (iceberg_level|int) is gt 3 %}
    local stratum 15
        {% else %}
```

## Info

BlueBanquise 1.3-RC2
CentOS Linux release 7.8.2003 (Core)
ansible 2.9.10
python 2.7.5
python-jinja2 2.7.2-4

## Workaround

```
# git diff
diff --git a/roles/core/time/templates/chrony.conf.j2 b/roles/core/time/templates/chrony.conf.j2
index 4503fd6..fd64f8b 100644
--- a/roles/core/time/templates/chrony.conf.j2
+++ b/roles/core/time/templates/chrony.conf.j2
@@ -61,7 +61,7 @@ allow {{ networks[nic.network]['subnet'] }}/{{ networks[nic.network]['prefix'] }
 # Serve time even if not synchronized to a time source.
 # Fairly unreliable time source
   {% if iceberg_level is defined and iceberg_level is not none %}
-    {% if (iceberg_level|int) is gt 3 %}
+    {% if (iceberg_level|int) > 3 %}
 local stratum 15
     {% else %}
 local stratum {{ 12+(iceberg_level|int) }}
```

## Note

It highlights an issue with the CI tests which run ansible from ubuntu-latest, with Python 3.

PR #345, opened 2020-08-27
fix: use '>' instead of 'gt' in j2 templates

btravouillon:
jinja2 2.7 does not support test 'is gt'.

Fixes: #329


oxedions (approved):
Dear @btravouillon 

Thanks for this PR.

Merging 🙂

PR #347, opened 2020-08-27
Added vars for the playbook of prometheus_client and prometheus_server

sobrase:
added start_services and enable_services because if the two vars are not set, the playbooks prometheus_client and prometheus_server will crash


btravouillon:
Can you update your commit message with a shorter subject line? (https://chris.beams.io/posts/git-commit/#limit-50)

PR #317, opened 2020-08-18
Revert "ci: do not install ansible 2.9.12"

btravouillon:
This reverts commit ac1b811c5c2852d5f804bc65d994d95b5deebc28.

Draft PR to troubleshoot the issues with CI after new ansible, molecule and ansible-lint releases.

PR #349, opened 2020-08-27
fix: define mode for directories

btravouillon:
Spotted by ansible-lint's rule E208. (#317)

Set mode 750 for /var/log/{conman,rsyslog}.
Skip the test for the nfs_client mount point.


oxedions (approved):
Dear @btravouillon 

Thanks a lot for this PR 😊 

Merging.

PR #350, opened 2020-08-28
fix: enable SELinux daemons_use_tty for conman

btravouillon:
conman service needs this tunable when SELinux is enabled on the host.


oxedions (approved):
Dear @btravouillon 

Thank you for this PR.

This is ok for me. Merging.

PR #352, opened 2020-08-28
fix: stop firewall service if ep_firewall = False

btravouillon:
Identified during testing of 1.3-RC2: after enabling the firewall for some tests, I noticed a new run of my playbook which includes role firewall did not disable the service.
If ep_firewall is False, do not omit the state but ensure the service is stopped.

Note: I turned ep_firewall to True to enable the firewall for my test.


oxedions (approved):
Dear @btravouillon 

Yes, I see your POV. This is ok for me.

Merging

PR #353, opened 2020-08-29
fix: update bluebanquise.spec for 1.3

btravouillon:
Install new internal inventory.
Add python requirements for Ansible filters.
Add CHANGLOG.md to documentation in RPM spec file.


oxedions (approved):
Ok make sense to add these packages as requirements.

Merging 🙂

PR #354, opened 2020-08-31
Fix: Add equipment_profiles ability to diskless chroot

oxedions:
Current documentation did not allowed to chose an equipment profile for the diskless image.
Also simplified the ansible-playbook execution line.


btravouillon (approved):
Looks good to me.

Issue #288, opened 2020-07-21
Repositories client role should handle both repositories formats for Centos


oxedions:
repositories_client role in core do not handle both repositories formats, aka:
```

  repositories:
    - os
    - bluebanquise
    - name: epel
      baseurl: 'https://dl.fedoraproject.org/pub/epel/$releasever/$basearch/'
      proxy: 'https://proxy:8080'
```

In the disable distribution repositories section of the task.

btravouillon:
https://github.com/bluebanquise/bluebanquise/pull/285#discussion_r458366763

btravouillon:
The bug occurs when `os` repo is not a string:
```
repositories:
  - bluebanquise
  - { name: 'os', gpgcheck: True }
```

PR #368, opened 2020-09-14
fix: disable distro repos even when 'os' is a dict

btravouillon:
See #288 

Also add new molecule tests for the role repositories_client with several configuration combination.


oxedions (approved):
Hello @btravouillon 

This is ok for me, thanks a lot for this PR 😊

PR #378, opened 2020-09-21
fix: don't create ini file if absent

btravouillon:
The role repositories_client can change the default repository
configuration for CentOS with the ini_file module. This should only
apply if the file already exists. If it does not exist, do not create the
file.

ini_file defaults to `create: yes`, which makes the CI fail because the `mode` is not defined.


oxedions (approved):
Dear @btravouillon 

Thank you for this PR.

Merging

Issue #296, opened 2020-07-28
disklessset does not exit gracefully when failing


btravouillon:
# Description

While reproducing an issue reported by a user with disklessset, I noticed the tool does not exit gracefully.
In the process, it creates and mounts a live image but does not umount it in case of failure in the process of image creation, leading to orphan files and eventually no space left in the filesystem.

# Info

OS: CentOS 8.1
Release: tag `1.3-rc1`

# Reproducer

```
[root@management1 bluebanquise]# df -h /root/diskless
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1        10G  6.3G  4.7G  64% /

[root@management1 bluebanquise]# disklessset
BlueBanquise Diskless manager
 1 - List available kernels
 2 - Generate a new initramfs
 3 - Generate a new diskless image
 4 - Manage/list existing diskless images
 5 - Remove a diskless image
-->: 3

Starting new image creation phase.
Many questions will be asked, a recap will be provided before starting procedure.

Select image type:
 1 - nfs
 2 - livenet
-->: 2
[INFO] Loading kernels from /var/www/html/preboot_execution_environment/diskless/kernels/

Select kernel:
 1 - vmlinuz-4.18.0-147.3.1.el8_1.x86_64
-->: 1
Please enter image name ?
-->: test1
Please enter clear root password of the new image: test1
[INFO] Entering livenet dedicated part.
Please select livenet image generation profile:
 1 - Standard: core (~1.2Gb)
 2 - Small: openssh, dnf and NetworkManager (~248Mb)
 3 - Minimal: openssh only (~129Mb)
-->: 3
Please choose image size, considering /1000: 2M for 2Gb, 600K for 600Mb, etc:
-->: 2M
Do you want to create a new NFS image with the following parameters:
  Image name:           test1
  Kernel version:       vmlinuz-4.18.0-147.3.1.el8_1.x86_64
  Root password:        test1
  Image profile:        3
  Image size /1000:     2M
Confirm ? Enter yes or no: yes
[INFO] Cleaning and creating image folders.
[INFO] Generating new ipxe boot file.
[INFO] Creating empty image file, format and mount it.
[...]
[something goes wrong]
Traceback (most recent call last):
  File "/usr/bin/disklessset", line 332, in <module>
    with open('/mnt/etc/shadow') as ff:
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/etc/shadow'
```
[...]

PR #379, opened 2020-09-22
fix: force to enter integer in select_from_list()

btravouillon:
select_from_list() must return an integer from the user selection, in
the range 1..len(list_from). Any other value is not valid. The function
will loop recursively until the user enter a valid value or exit the
programm.

Parameter index_modifier defaults to -1 since every call to the function
is using this value.

Avoid useless type casting int(str(int(x))).

See: #296


oxedions (commented):
Dear @btravouillon 

Thank you for this PR.

Merging

PR #381, opened 2020-09-22
fix: check if diskless image_metadata.yml exists

btravouillon:
The image folder is created at the begining of the image creation
process while the image_metadata.yml file is created at the end. When
the image creation process fails, the directory exists but the metadata
file does not exist.

Check if the metadata file exists before read. Otherwise, print a
warning to inform the user the image is incomplete.

See: #296


oxedions (approved):
Dear @btravouillon 

Thank you for this PR.

Merging

PR #388, opened 2020-09-23
fix: ensure ep_ipxe_embed is defined before use

btravouillon:
If not defined, it will default to 'standard'.


bouriquet (approved):

oxedions:
Isn't ep_ipxe_embed always defined ? It is in group_vars/all/equipment_all/equipment_profile.yml so should always be present.

btravouillon:
This was introduced with 1.2.0, users of 1.1 won't have this variable defined in their inventory.

oxedions (approved):
Ok for me then 😊

Issue #360, opened 2020-09-03
pxe_stack - menu.ipxe miss the grub shell entry point


oxedions:
Like for ipxe, a grub2 shell image was provided and entry is missing in the menu.ipxe (I just forgot it during devs, but image is present on the system, provided by ipxe-bluebanquise rpms).

This image was here to help sysadmin to debug system boot related issues or failing of the grub2 automated image (EFI disk boot).

PR #389, opened 2020-09-24
Fix: pxe_stack - add grub2 shell in menu.ipxe

oxedions:
Add the grub2 shell entry in the menu.
Was present in the rpm, but missing here.

Related to https://github.com/bluebanquise/bluebanquise/issues/360

### DIFF
--- a/roles/core/pxe_stack/templates/menu.ipxe.j2
+++ b/roles/core/pxe_stack/templates/menu.ipxe.j2
@@ -40,7 +40,7 @@ item bootdeployclone      Deploy already made clone
 #item memtest              Memtest (To check Memory)
 item --gap                Other
 item startshell           iPXE shell (Go to an iPXE shell)
-item startgrubshell       Grub2 shell (Go to a Grub2 shell)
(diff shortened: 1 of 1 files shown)


hmeAtos (dismissed):
ok for me

roles/core/pxe_stack/templates/menu.ipxe.j2
```diff
@@ -40,6 +40,7 @@ item bootdeployclone      Deploy already made clone
 #item memtest              Memtest (To check Memory)
 item --gap                Other
 item startshell           iPXE shell (Go to an iPXE shell)
+item startgrubshell       Grub2 shell (Go to a Grub2 shell)
```
btravouillon, line 43: Can you add information this is for EFI only?

btravouillon (commented):
The grub2_shell.img seems to be EFI only.

roles/core/pxe_stack/templates/menu.ipxe.j2
oxedions: Doing it right now. It is indeed EFI only.

btravouillon (approved):
Looks good to me.

PR #390, opened 2020-09-24
Feat: Add Lmod role

oxedions:
Add Lmod and configure custom path if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.4.0
+## 1.4.0 - Next-release
 
 ### Major changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 1.4.0
+
+### Major changes
+
+#### New roles
+
+  - addons/Lmod: allow to install Lmod and specify custom modulefiles path
+
+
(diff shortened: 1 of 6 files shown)


roles/addons/Lmod/vars/RedHat.yml
```diff
@@ -0,0 +1,3 @@
+---
+Lmod_packages_to_install:
```
strus38, line 2: why a capital lettre for a variable?

oxedions: Hello !
Well, because the tool itself is everywhere using capital L (doc, spec, code, etc). So I tried to comply with that 🙂

strus38:
Ok... I don't like capital letters, it sounds too much like windows :-)

btravouillon (dismissed):
Looks good to me. We may need to add some CI tests.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 1.4.0
```
btravouillon, line 3: On a second thought, should you replace 1.4.0 with `devel` or `next-release`?

btravouillon: Then we will update the version to `1.4.0-RCx`, then `1.4.0` once ready to release.

btravouillon (approved):

PR #394, opened 2020-09-29
fix: missing packages in conman role

bouriquet:
seboolean module requires python selinux libs

### DIFF
--- a/roles/core/conman/tasks/main.yml
+++ b/roles/core/conman/tasks/main.yml
@@ -21,11 +21,9 @@
   tags:
     - internal
 
-- name: package █ Install conman and ipmitool
+- name: package █ Install packages
   package:
-    name:
-      - conman
-      - ipmitool
+    name: "{{ conman_packages_to_install }}"
     state: present
   tags:
     - package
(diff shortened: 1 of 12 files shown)


hmeAtos (dismissed):
good catch !
ok for me

roles/core/conman/tasks/main.yml
```diff
@@ -21,11 +21,9 @@
   tags:
     - internal
 
-- name: package █ Install conman and ipmitool
+- name: package █ Install packages
```
btravouillon, line 24: Most roles use `{{ rolename_packages_to_install | join(' ') }}`

btravouillon (commented):
Indeed, nice catch.
Could you also update `nfs_client` and `openldap_server` that may have the same issue?
```
$ grep -r seboolean: roles/
roles/addons/openldap_server/tasks/main.yml:  seboolean:
roles/core/nfs_client/tasks/main.yml:  seboolean:
roles/core/conman/tasks/main.yml:  seboolean:
```

roles/core/conman/tasks/main.yml
bouriquet: done

btravouillon (approved):
Looks good to me.

oxedions (approved):

PR #398, opened 2020-10-04
ci: run molecule test with tox

btravouillon:
[tox](https://tox.readthedocs.io/en/latest/) helps to run `molecule test` in several environments.

This PR defines a combination of test environments:
-  Python 2.7 and Python 3
-  Ansible 2.8, 2.9 and 2.10

The purpose is to expand the test coverage.

Additional changes:
- Do not run ansible-lint and flake8 in molecule
    
    The static code analysis CI tests already run those tools with GitHub
    Actions for each PR or push to master. With the current tox
    configuration, the CI would run these 6x for each role, which is
    useless.

- Clean and fix some verify tests.

- Serialize the molecule converge for role time

### DIFF
--- a/.github/workflows/molecule-actions-addons-clustershell.yml
+++ b/.github/workflows/molecule-actions-addons-clustershell.yml
@@ -35,7 +35,7 @@ jobs:
         uses: actions/setup-python@v1
       - name: Install packages
         run: |
-          pip install 'molecule[docker]>3' ansible-lint flake8
+          pip install tox
       - name: Run molecule test for addons/clustershell
         run: |
-          cd roles/addons/clustershell && molecule test
+          cd roles/addons/clustershell && tox
--- a/.github/workflows/molecule-actions-addons-grafana.yml
+++ b/.github/workflows/molecule-actions-addons-grafana.yml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/setup-python@v1
       - name: Install packages
         run: |
-          pip install 'molecule[docker]>3' ansible-lint flake8
+          pip install tox
       - name: Run molecule test for addons/grafana
         run: |
-          cd roles/addons/grafana && molecule test
+          cd roles/addons/grafana && tox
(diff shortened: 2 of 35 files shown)


btravouillon:
core/time unit testing is failing because of idempotence issue:
```
ERROR: Idempotence test failed because of the following tasks:
* [client-ext-servers] => time : timezone █ Set system to Europe/Paris timezone
* [client-ext-pool] => time : timezone █ Set system to Europe/Paris timezone
```

This is an error I saw from time to time in GH Actions. It seems to be related to the testing environment. Previously, a new run of the job was sufficient to get the green light. With tox...6x tests per OS... well... I tried several time to re-run the job, but statistically one env will fail.
Unfortunately I can't reproduce this locally (Debian 10), the molecule tests always succeed.

btravouillon:
I was able to reproduce in an Ubuntu 18.04 VM. First observations make me believe there is a race when accessing the cgroup from all the privileged containers at the same time.

In commit 71e199a, I enable the serialization of the converge playbook to run it one host at a time.

With this change, the test is green. However it took ~1h15 to complete: https://github.com/bluebanquise/bluebanquise/actions/runs/315669729

btravouillon:
Green in 40 minutes with `serial: 3`. :grin: 
Let's keep this value. If the test should fail again, we might consider to decrease the value.

oxedions:
Being curious 🙂 

What is the meaning of:

```
- name: Converge
  hosts: all
  serial: 3
```

Only three hosts at a time ?

btravouillon:
> Only three hosts at a time ?

Yes. See [Setting the batch size with serial](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#setting-the-batch-size-with-serial). Previously the converge playbook was running with the 6 containers/hosts in parallel.

oxedions:
Thanks a lot.
I will be honest, this is something I do not master, so my review is only partial, and based on the impact on the stack itself.

oxedions (approved):
Thanks a lot for this massive PR. Nice work !

PR #386, opened 2020-09-23
fix: wrong images selinux context in disklessset

bouriquet:
Hi

disklessset image generation leaves images (squash + boot.ipxe) files with the wrong selinux context. Thoses files cannot be served when selinux is enabled on managements.

PR #404, opened 2020-10-08
fix: ensure selinux type is correct on diskless directories

bouriquet:
#386


btravouillon (approved):
I like it! Thanks!

oxedions (approved):
Dear @bouriquet 

Thanks a lot for this PR 😊 

Ok for me, merging.

PR #405, opened 2020-10-09
Fix: documentation - corrections based on feedback

oxedions:
Remove disabling firewall since stack now handles it.
Increase minimum hdd requirement.

Too much people fall in this firewall issue.


btravouillon (approved):
Looks good to me. 👍

PR #406, opened 2020-10-09
Fix: hosts_file - Prevent massive empty lines generation

oxedions:
Prevent many empty lines when hosts are defined in the inventory, but without network_interfaces.

### DIFF
--- a/roles/core/hosts_file/templates/hosts.j2
+++ b/roles/core/hosts_file/templates/hosts.j2
@@ -77,9 +77,7 @@
 ## Internal hosts
 
 {% set current_iceberg = j2_current_iceberg %}
-{% for host in range %}
-{{ write_host(host,hostvars[host],current_iceberg) }}
-{% endfor %}
+{% for host in range %}{{ write_host(host,hostvars[host],current_iceberg) }}{% endfor %}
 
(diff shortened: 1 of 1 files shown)


hmeAtos (dismissed):
Good, ok for me

roles/core/hosts_file/templates/hosts.j2
```diff
@@ -77,9 +77,7 @@
 ## Internal hosts
 
 {% set current_iceberg = j2_current_iceberg %}
-{% for host in range %}
-{{ write_host(host,hostvars[host],current_iceberg) }}
-{% endfor %}
+{% for host in range %}{{ write_host(host,hostvars[host],current_iceberg) }}{% endfor %}
```
btravouillon, line 80: We loose in readability. Can you consider the change below?

```suggestion
{% for host in range -%}
{{ write_host(host,hostvars[host],current_iceberg) }}
{%- endfor %}
```

oxedions: Sure, let me test that.

oxedions: ok !

PR #411, opened 2020-10-12
[diskless] The time role isn't functional for a custom diskless image

hmeAtos:
Hello,

The timezone module isn't designed to work in a chroot environment, so we have to add the identify tag in task role.

```
RUNNING HANDLER [log_client : Restart rsyslog services] 
*************************************************************
Thursday 01 October 2020 02:53:57 +0200 (0:00:00.338) 0:00:12.768 ******
ok: [/var/tmp/diskless/workdir/dkl_last_kernel_custom/mnt] => (item=rsyslog)
[WARNING]: Target is a chroot or systemd is offline. This can lead to false positives or prevent the init system tools from working.

TASK [log_client : Manage services state] 
*******************************************
```

Regards

### DIFF
--- a/roles/addons/diskless/readme.rst
+++ b/roles/addons/diskless/readme.rst
@@ -303,6 +303,9 @@ Example Playbook
       - diskless
 
 
+Once the node is started, run your playbook with your roles.
+It is important to synchronize your node's time by running the role time.
+
 To be done
 ^^^^^^^^^^
 
--- a/roles/addons/diskless/readme.rst
+++ b/roles/addons/diskless/readme.rst
@@ -304,7 +304,7 @@ Example Playbook
 
 
 Once the node is started, run your playbook with your roles.
-It is important to synchronize your node's time by running the role time.
+It is important to synchronize your node's time by running the time role.
 
 To be done
 ^^^^^^^^^^
--- a/roles/core/time/tasks/main.yml
+++ b/roles/core/time/tasks/main.yml
@@ -56,8 +56,6 @@
 - name: "timezone █ Set system to {{ time_zone }} timezone"
   timezone:
     name: "{{ time_zone }}"
-  tags:
-    - identify
 
 - meta: flush_handlers
 
--- a/roles/core/time/tasks/main.yml
+++ b/roles/core/time/tasks/main.yml
@@ -56,6 +56,8 @@
 - name: "timezone █ Set system to {{ time_zone }} timezone"
   timezone:
     name: "{{ time_zone }}"
+  tags:
+    - identify
 
 - meta: flush_handlers
 


btravouillon (commented):
From roles/addons/diskless/readme.rst:

> The `--skip-tags identify` prevents hostname and static ip to be set, since the image should be generic for multiple hosts.

It seems to be wrong usage here since the timezone does not identify a node.

Moreover, the message in the description is about handler **log_client : Restart rsyslog services** while the change is about task **timezone █ Set system to {{ time_zone }} timezone**. What is the link between both?

hmeAtos:
> From roles/addons/diskless/readme.rst:
> 
> > The `--skip-tags identify` prevents hostname and static ip to be set, since the image should be generic for multiple hosts.
> 
> It seems to be wrong usage here since the timezone does not identify a node.
> 
> Moreover, the message in the description is about handler **log_client : Restart rsyslog services** while the change is about task **timezone █ Set system to {{ time_zone }} timezone**. What is the link between both?

Indeed you're right, I'm fixing that.
Thanks

oxedions:
I think timedatectl (which cannot be run inside a chroot) only do a simple link. So maybe role could be enhance to detect chroot environment, and if chroot, rely on creating a link instead of the standard timezone module ?

btravouillon:
The description of this PR is still about `log_client`, I can't find the link between the title and the description.

I'm not keen to hack the role to detect if the task is run through a chroot connection and to apply some ugly workaround. Either it works with the timezone module, either we replace the module if it does not cover our use cases.

That being said, the `time_zone` is a host property, it is preferable to configure it at boot, not in the image. A use case could be 2 clusters, in 2 different time zones, using the same diskless images. One can either configure the TZ with the role `time` at boot, or enable the tcode option of DHCP. (dhcp-options(5), [RFC 4833](https://tools.ietf.org/html/rfc4833))

oxedions:
I understand that.

Then I suggest we keep with this documentation. Users will have to execute playbook after boot of nodes, using `-t identify,time`.

btravouillon (approved):
Looks good to me. 👍

oxedions (approved):

oxedions:
Cool ! Then merging.

btravouillon:
Reminder for the team @bluebanquise/bluebanquise: please ensure your commit messages are useful!

See the current git log:
git_log

This commit message is bad, it does not describe the change. Moreover, despite I wrote **twice** ([here](https://github.com/bluebanquise/bluebanquise/pull/411#pullrequestreview-506676682) and [there](https://github.com/bluebanquise/bluebanquise/pull/411#issuecomment-708661099)) the title is about **time** and the description is about **log_client**, nothing was changed. Now our log history is dirty... not cool 😞 

Ask yourself: what useful information can I get from this commit message?

Read: https://www.janmeppe.com/blog/good-commit-messages/

hmeAtos:
> Reminder for the team @bluebanquise/bluebanquise: please ensure your commit messages are useful!
> 
> See the current git log:
> git_log
> 
> This commit message is bad, it does not describe the change. Moreover, despite I wrote **twice** ([here](https://github.com/bluebanquise/bluebanquise/pull/411#pullrequestreview-506676682) and [there](https://github.com/bluebanquise/bluebanquise/pull/411#issuecomment-708661099)) the title is about **time** and the description is about **log_client**, nothing was changed. Now our log history is dirty... not cool 😞
> 
> Ask yourself: what useful information can I get from this commit message?
> 
> Read: https://www.janmeppe.com/blog/good-commit-messages/

It's noted thank you.

PR #412, opened 2020-10-12
Add molecule tests for nfs_server and nfs_client

osmocl:
Molecule tests for:

- nfs_server
- nfs_client

Note that it is not possible to run properly a nfs-server in a docker container.

Add new variable (nfs_client_directories_state) to allow all possible mount values in state parameter.

Variable name `nfs_client_directories_state`

variable values: 

- mounted
- absent
- present
- unmounted
- remounted


osmocl:
Solving issues with checks in **core/nfs_client unit testing / centos:7**

.github/TEST_MATRIX.md
```diff
@@ -15,6 +15,8 @@ Molecule unit tests
 |![core/firewall unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/firewall%20unit%20testing/badge.svg)                       | x | x |   |
 |![core/log_client unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/log_client%20unit%20testing/badge.svg)                   | x | x | x |
 |![core/log_server unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/log_server%20unit%20testing/badge.svg)                   | x | x | x |
+|![core/nfs_client unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/nfs_client%20unit%20testing/badge.svg)                   | x | x | x |
```
btravouillon, line 18: Can you uncheck OpenSUSE?

roles/core/nfs_client/molecule/default/molecule.yml
```diff
@@ -0,0 +1,26 @@
...
+    capabilities:
+      - "SYS_ADMIN"
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
+    pre_build_image: true
+provisioner:
+  name: ansible
+  ansible_args:
+    - -e mount_exported_dirs=false
```
btravouillon, line 20: We are used to define global vars in converge.yml ([example](https://github.com/bluebanquise/bluebanquise/blob/master/roles/core/log_client/molecule/default/converge.yml)).

roles/core/nfs_client/molecule/default/molecule.yml
```diff
@@ -0,0 +1,26 @@
...
+  inventory:
+    links:
+      group_vars: ../../../../../resources/examples/simple_cluster/inventory/group_vars
```
btravouillon, line 23: Interesting usage of an example inventory. For molecule, I prefer not to rely on the examples inventory since they only cover a single use case, even if we can override any variable in the molecule files.
@oxedions @strus38 any comment on this?

roles/core/nfs_client/molecule/default/verify.yml
```diff
@@ -0,0 +1,56 @@
...
+      register: register_stat
+
+    - name: Assert dir /etc/software exist
+      assert:
+        that: 
+          - "register_stat.stat.exists"
+          - "register_stat.stat.isdir"
+
+    - name: Retrieve /opt/software in /etc/fstab file
+      lineinfile:
```
btravouillon, line 29: If I remember correctly this module modifies the file.
I would take an approach similar to [the verify of the role time](https://github.com/bluebanquise/bluebanquise/blob/0cfa48ba0e09991753cdbe98f3e1176ac808a295/roles/core/time/molecule/default/verify.yml#L30):
 - read /etc/fstab once
 - assert a line exists with regex_search

roles/core/nfs_client/vars/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
-nfs_client_role_version: 1.0.3
+nfs_client_role_version: 1.0.4
+
+mount_exported_dirs: true # true=mounted, false=present
```
btravouillon, line 4: Please use prefix `nfs_client_` for variables in this role. If this is a user variable, please move it to `defaults/main.yml`. Variables in `vars/main.yml` are internal to the roles.

Also, instead of a boolean, I would allow to define any state supported by the module `mount`. Is there a reason to only allow mounted/present? (which is still better than enforced `mounted`!! :wink:)

roles/core/nfs_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,68 @@
...
+      stat:
+        path: /etc/exports
+      register: register_stat
+
+    - name: Assert file /etc/exports exist
+      assert:
+        that: "register_stat.stat.exists"
+
+    - name: Retrieve /opt/software in /etc/exports file
+      lineinfile:
```
btravouillon, line 39: Same comment as for role nfs_server about lineinfile.

btravouillon (commented):
Hi @osmocl ,
Thank you very much for this PR. A few comments inline.

roles/core/nfs_client/molecule/default/verify.yml
btravouillon: On second read, I understand the test:

- check for the line with regexp
- remove the line if present
- check mode (let the file untouched)
- test the result

It is somewhat different than what I saw in the past. It works.

.github/TEST_MATRIX.md
osmocl: Yes, I am going to correct that, I am not checking OpenSUSE.

roles/core/nfs_client/molecule/default/molecule.yml
osmocl: Yes, I used to declare variables in converge.yml too, however here it is not working.
I have declared directly the variable like an extra var to have the highest priority.
I will double chek that.

roles/core/nfs_client/molecule/default/molecule.yml
osmocl: If we provides examples, why we do not use them ? 
I love this approach because It is cleaner to read.
Besides, using examples allow us to catch errors automatically for example if a variable name changes in `nfs.yml` file.

roles/core/nfs_client/molecule/default/verify.yml
osmocl: Yes, I am using the check_mode here.

roles/core/nfs_client/vars/main.yml
osmocl: Ok, I will prefix and move the variables.
I have never used the states: `unmounted ` or `absent `with ansible.
Anyway, by default the role `nfs_client ` use the `mounted` option, this is not changing here.
I propose to use the `present` option uniquely to be able to run this unitary tests, this does not impact the role installation.

osmocl: Corrected in commit f28b555

roles/core/nfs_client/molecule/default/molecule.yml
osmocl: Corrected in commit f28b555

.github/TEST_MATRIX.md
osmocl: Corrected in commit f28b555

roles/core/nfs_client/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+nfs_client_role_version: 1.0.4
+
```
btravouillon: This is a variable internal to the role, already in `vars/main.yml`. Can you remove these 2 lines?

roles/core/nfs_client/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
...
+nfs_client_mount_exported_dirs: true # true=mounted, false=present
```
btravouillon, line 2: While I understand the role previously enforced `mounted`, I would be extra careful with such boolean variables. Any future RFE to add support for a third value will lead to a breaking change. We had the same issue with ep_access_controle for 1.3.
I still prefer to allow explicit mounted/present/whatever supported state.
@bluebanquise/bluebanquise: I would like comments from other reviewers.

roles/core/nfs_client/defaults/main.yml
btravouillon: OK, GitHub displayed these lines in the review while you removed those in a later commit... My bad.

roles/core/nfs_client/defaults/main.yml
osmocl: I just allowed all possible mount values in state parameter, commit: 4dd6a48

osmocl:
Static analysis is failing here too.
I propose to add the E301 to be excluded in `bluebanquise/.github/workflows/static_analysis.yml`. What do you think ?

```
      - name: Run ansible linter
        run: ansible-lint -v -x 204 -x 301
```

roles/core/nfs_client/molecule/default/molecule.yml
oxedions: This is interesting because it kind of validate also the examples. If we update the examples and "break" something, then all tests will fail.
Are we able to do one first "pass" with the example inventories (both ?) and then optionally do specific "passes" if we need to cover role specific use cases ?

roles/core/nfs_client/defaults/main.yml
oxedions: 👍 +1, all values should be available.

oxedions: So maybe rename it to: `nfs_client_exported_directories_state `? Since this refer to `state `in the module.

roles/core/nfs_client/molecule/default/molecule.yml
btravouillon: > using examples allow us to catch errors automatically for example if a variable name changes in nfs.yml file.
> [...]
> If we update the examples and "break" something, then all tests will fail.

This is exactly why I don't want to include these examples to the role tests. We are testing the roles, not the examples. Moreover, examples must remain examples. If we use these in the tests, they will become de facto standards.

> Are we able to do one first "pass" with the example inventories (both ?) and then optionally do specific "passes" if we need to cover role specific use cases ?

It might be possible to create molecule tests dedicated to the examples that would run the example playbooks along with the example inventories.

btravouillon:
> Static analysis is failing here too.
> I propose to add the E301 to be excluded in `bluebanquise/.github/workflows/static_analysis.yml`. What do you think ?

I don't want to exclude a rule for the whole project because it is failing in new code. I'll take a look at the failures.

btravouillon:
> Static analysis is failing here too.

The command does not bring any change to the system, it only fetches information. You can safely add a `changed_when: False`.

```
$ git diff
diff --git a/roles/core/nfs_server/molecule/default/verify.yml b/roles/core/nfs_server/molecule/default/verify.yml
index 7d2dd35..91dc33c 100644
--- a/roles/core/nfs_server/molecule/default/verify.yml
+++ b/roles/core/nfs_server/molecule/default/verify.yml
@@ -13,6 +13,7 @@
     - name: Collect services of zone public
       command: firewall-cmd --zone=public --list-all
       register: firewall_cmd_result
+      changed_when: False
 
     - name: Check rpc-bind and nfs are in zone public
       assert:
```

oxedions:
@osmocl 
If possible, remember to also add an entry in the CHANGELOG.md since there is a change in the nfs_client role (new variable).

roles/core/nfs_client/molecule/default/molecule.yml
oxedions: >This is exactly why I don't want to include these examples to the role tests. We are testing the roles, not the examples. Moreover, examples must remain examples. If we use these in the tests, they will become de facto standards.

I can understand that. Then lets keep it the way it is.

roles/core/nfs_server/readme.rst
```diff
@@ -60,6 +60,7 @@ Packages installed:
 Changelog
 ^^^^^^^^^
 
+* 1.0.4: Added molecule tests. osmocl <osmocl@osmo.cl>
```
btravouillon, line 63: Please update the changelog message to reflect the change for end-users (new variable)

roles/core/nfs_client/molecule/default/molecule.yml
osmocl: Ok guys, fair enough, I am taking in count this discussion, I will rework all the `molecule.yml` of my PR's.
Thank you !

osmocl: Inventory part revamped, commit : 4dd4ed5

roles/core/nfs_client/defaults/main.yml
osmocl: Variable renamed as requested, commit : 4dd4ed5

roles/core/nfs_server/readme.rst
osmocl: Changelog updated in commit : 4dd4ed5

roles/core/nfs_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,69 @@
...
+        path: /etc/exports
+        regexp: "^/opt/software"
+        state: absent
+      check_mode: yes
+      register: export_sw_result
```
btravouillon, line 45: Can you add `changed_when: false` so the verify won't report changes in the output?

roles/core/nfs_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,69 @@
...
+      assert:
+        that: export_sw_result.found
+
+    - name: Retrieve /home in /etc/exports file
+      lineinfile:
+        path: /etc/exports
+        regexp: "^/home"
+        state: absent
+      check_mode: yes
+      register: export_home_result
```
btravouillon, line 57: Can you add `changed_when: false` so the verify won't report changes in the output?

roles/core/nfs_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,69 @@
...
+      register: export_sw_result
+
+    - name: Check /etc/exports file contains /opt/software line
+      assert:
+        that: export_sw_result.found
+
+    - name: Retrieve /home in /etc/exports file
+      lineinfile:
+        path: /etc/exports
+        regexp: "^/home"
```
btravouillon, line 54: Not only the role set the export point in /etc/exports, it also defines the list of clients and the export options in parenthesis. Can you check all these values here to ensure a change to any variable won't break?

roles/core/nfs_server/molecule/default/verify.yml
```diff
@@ -0,0 +1,69 @@
...
+      register: register_stat
+
+    - name: Assert file /etc/exports exist
+      assert:
+        that: "register_stat.stat.exists"
+
+    - name: Retrieve /opt/software in /etc/exports file
+      lineinfile:
+        path: /etc/exports
+        regexp: "^/opt/software"
```
btravouillon, line 42: Not only the role set the export point in /etc/exports, it also defines the list of clients and the export options in parenthesis. Can you check all these values here to ensure a change to any variable won't break?

btravouillon (changes requested):

btravouillon:
Hello @osmocl ,

#398 merged yesterday, the CI now use tox to test several version of ansible. Can you update all your PR to use tox?
This is a straightforward change: you need to update the GH Actions workflow file for the job, pip install tox instead of molecule + deps, and run tox instead of molecule. (see 146f09da for an example)
Tox will install molecule and all the deps in each virtualenv, then execute the molecule test.

Thanks.

roles/core/nfs_client/molecule/default/verify.yml
```diff
@@ -0,0 +1,56 @@
...
+        path: /etc/fstab
+        regexp: "/opt/software nfs ro,intr,nfsvers=4.2,bg"
+        state: absent
+      check_mode: yes
+      register: export_sw_result
```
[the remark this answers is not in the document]
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_client/molecule/default/verify.yml
```diff
@@ -0,0 +1,56 @@
...
+      assert:
+        that: export_sw_result.found
+
+    - name: Retrieve /home in /etc/fstab file
+      lineinfile:
+        path: /etc/fstab
+        regexp: "/home nfs rw,intr,rsize=32768,wsize=32768,nfsvers=4.2,bg"
+        state: absent
+      check_mode: yes
+      register: export_home_result
```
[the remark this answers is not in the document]
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_client/molecule/default/verify.yml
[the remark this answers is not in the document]
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_client/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+nfs_client_directories_state: mounted # possible values: mounted, absent, present, unmounted, remounted
```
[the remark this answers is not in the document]
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_client/tasks/main.yml
```diff
@@ -63,7 +63,7 @@
     src: "{{ item.0.server }}{% if item.0.take_over_network is defined and item.0.take_over_network is not none %}-{{ item.0.take_over_network }}{% endif %}:{{ item.0.export }}"
     fstype: nfs
     opts: "{{ item.0.mount_arguments }}"
-    state: mounted
+    state: "{{ nfs_client_directories_state | default('mounted') }}"
```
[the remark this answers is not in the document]
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

roles/core/nfs_server/molecule/default/verify.yml
osmocl: Done in commit: 5cb76a02a40198f54bb9aab67f55e46f1f43850b

btravouillon (approved):
Looks good to me. 👍
Thank you @osmocl.

oxedions (approved):
Many thanks to both of you !

It is ok for me.

Merging 🙂

PR #413, opened 2020-10-14
Add unitary molecule tests for dhcp_server role

osmocl:
- Assing static ip in docker containers
- Using a fake host declared in` molecule.yml`
- Run unitary tests


.github/TEST_MATRIX.md
```diff
@@ -12,6 +12,7 @@ Molecule unit tests
 |![core/display_tuning unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/display_tuning%20unit%20testing/badge.svg)           | x | x |   |
 |![core/dns_client unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/dns_client%20unit%20testing/badge.svg)                   | x | x | x |
 |![core/dns_server unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/dns_server%20unit%20testing/badge.svg)                   | x | x |   |
+|![core/dhcp_server unit testing](https://github.com/bluebanquise/bluebanquise/workflows/core/dhcp_server%20unit%20testing/badge.svg)                 | x |   |   |
```
btravouillon, line 15: Can you please keep the roles alphabetically sorted?

.github/workflows/molecule-actions-core-dhcp_server.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  molecule_test:
+    name: ${{ matrix.distro }} - molecule test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        distro:
+          - centos:8
```
btravouillon, line 23: Is there any failure with CentOS 7?

roles/core/dhcp_server/molecule/default/molecule.yml
```diff
@@ -0,0 +1,100 @@
...
+    networks:
+      - name: bbnet
+        ipv4_address: '10.10.2.1'    
+
+provisioner:
+  name: ansible
+  inventory:
+    links:
+      hosts: ../../../../../resources/examples/simple_cluster/inventory/cluster/nodes
+      group_vars: ../../../../../resources/examples/simple_cluster/inventory/group_vars
```
btravouillon, line 97: Please don't use the examples in the tests. Explicitly define each requested variable/additional host instead.

roles/core/dhcp_server/readme.rst
```diff
@@ -116,6 +116,7 @@ Files generated:
 Changelog
 ^^^^^^^^^
 
+* 1.0.8: Add unitary molecule tests. Osmocl <osmocl@osmo.cl>
```
btravouillon, line 119: Please don't change the role version when there is no functional change or documentation improvement.

roles/core/dhcp_server/molecule/default/molecule.yml
```diff
@@ -0,0 +1,100 @@
...
+    docker_networks:
+      - name: bbnet
+        ipam_config:
+          - subnet: '10.10.0.0/16'
+            gateway: "10.10.0.254"
+    networks:
+      - name: bbnet
+        ipv4_address: "10.10.0.1"            
+
+  - name: c001
```
btravouillon, line 27: All prepare/converge/verifiy playbooks run on management1. What is the purpose of these additional containers?
If you need extra fake hosts, no need to spawn containers, you can add some to your inventory. (see [ssh_master](https://github.com/bluebanquise/bluebanquise/blob/0d0d2a08f3acea5a45215b4a70ad0268392061ba/roles/core/ssh_master/molecule/default/molecule.yml#L74))

btravouillon (commented):
Hi @osmocl ,
Thank you for this new test! A few comments inline.

roles/core/dhcp_server/molecule/default/molecule.yml
osmocl: I am injecting all inventories from `resources/examples/simple_cluster/inventory/cluster/nodes`. For now these containers are only reflecting the simple cluster architecture from the example.

roles/core/dhcp_server/readme.rst
osmocl: Ok I will remove it.

roles/core/dhcp_server/molecule/default/molecule.yml
osmocl: Sure I can rework it, but before I just want to understand why we cannot use the examples provided in BB, any specific reason ?
Why we need to rewrite every sigle inventory and needed vars if they already exist in examples ?
Let me know please if this approach is incorrect, because I am thinking to work in some others roles.

.github/workflows/molecule-actions-core-dhcp_server.yml
osmocl: Yes it was't running in my platform. I was getting several errors in the container, unfortunately I was't able to solve this.
For now only tests in cento8 are working. Do you have any idea about this error? 

Errror output :

`TASK [dhcp_server : package █ Install dhcp] ************************************
fatal: [management1]: FAILED! => {"changed": false, "changes": {"installed": ["dhcp"]}, "msg": "http://mirrors.vcea.wsu.edu/centos/7.8.2003/os/x86_64/Packages/dhcp-4.2.5-79.el7.centos.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: mirrors.vcea.wsu.edu; Unknown error\"\nTrying other mirror.\nhttp://mirrors.vcea.wsu.edu/centos/7.8.2003/os/x86_64/Packages/dhcp-libs-4.2.5-79.el7.centos.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: mirrors.vcea.wsu.edu; Unknown error\"\nTrying other mirror.\nhttp://mirror.centos.lax1.serverforge.org/7.8.2003/updates/x86_64/Packages/bind-export-libs-9.11.4-16.P2.el7_8.6.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: mirror.centos.lax1.serverforge.org; Unknown error\"\nTrying other mirror.\nhttp://mirrors.vcea.wsu.edu/centos/7.8.2003/os/x86_64/Packages/dhcp-common-4.2.5-79.el7.centos.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: mirrors.vcea.wsu.edu; Unknown error\"\nTrying other mirror.\nhttp://repo1.sea.innoscale.net/centos/7.8.2003/os/x86_64/Packages/dhcp-libs-4.2.5-79.el7.centos.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: repo1.sea.innoscale.net; Unknown error\"\nTrying other mirror.\nhttp://centos.s.uw.edu/centos/7.8.2003/os/x86_64/Packages/dhcp-4.2.5-79.el7.centos.x86_64.rpm: [Errno 14] curl#6 - \"Could not resolve host: centos.s.uw.edu; Unknown error\"\nTrying other mirror 
[...] `

.github/TEST_MATRIX.md
osmocl: Yes, I will correct it.

roles/core/dhcp_server/molecule/default/molecule.yml
btravouillon: I don't understand the purpose of these containers in the test. Why would we need to spawn these containers to test the role `dhcp_server`?

.github/workflows/molecule-actions-core-dhcp_server.yml
btravouillon: I'm having a hard time trying to reproduce this on my dev system (Debian 10.6)... The IP subnet used for the Docker network `bbnet` conflicts with my Vagrant setup (which uses a copy of the example inventory for dev/testing of playbooks). It would be nice to use a different subnet than 10.10.0.0/16 for developers who run a virtual environment for dev & integration tests and molecule for unit tests, all on the same host.

With a different subnet, the container `management1` loses access to the network after the start of the firewalld service, the system cannot resolve any address, which seems close to what you observe. However, I observe this with CentOS 8 only. With CentOS 7, the task `[dhcp_server : package █ Install dhcp]` succeeds. 😵

.github/TEST_MATRIX.md
osmocl: Corrected in commit: fd4d355

roles/core/dhcp_server/readme.rst
osmocl: Corrected in commit: fd4d355

roles/core/dhcp_server/molecule/default/molecule.yml
osmocl: I wanted to add some more advanced tests from clients to management, that is why I added extra containers.
Anyway, for now I believe that the number of tests that are being run in management are quite enough.
We will run more advanced functional tests in our own platform with kitchen.

Client containers removed in commit fd4d355

osmocl:
In my platform the check `E301 ` for ansible-lint is working perfectly in `skip_list ` and `warn_list`.
I dot not know why it is not working here.

btravouillon:
> In my platform the check `E301 ` for ansible-lint is working perfectly in `skip_list ` and `warn_list`.
> I dot not know why it is not working here.

ansible-lint runs at the root of the project, it won't look at roles/core/dhcp_server/.ansible-lint.

This is the same failure than in #412, with the same command. You can safely add `changed_when: False`.

roles/core/dhcp_server/molecule/default/molecule.yml
btravouillon: Great. I'm not against adding client/server tests, while it can be challenging with containers. If so, add the containers when they are effectively requested for the test. Thanks. :wink:

roles/core/dhcp_server/molecule/default/molecule.yml
btravouillon: We are testing the roles, not the examples. Examples must remain examples. If we use these in the tests, they will become de facto standards.

I like to keep each role independent, with all the requested data available in the molecule directory. It's easier to read and debug. All our molecule unit tests are defined this way.

However, this is an interesting use case to test the example inventories. If this is something of interest, I would prefer a dedicated molecule test, global to the project, for each example inventories and playbooks.

osmocl: I have removed the links, now I am declaring a fake host directly in `molecule.yml`
Corrected in commit: 892bd4f

roles/core/dhcp_server/molecule/default/molecule.yml
```diff
@@ -0,0 +1,75 @@
...
+    capabilities:
+      - "SYS_ADMIN"
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
+    pre_build_image: true
+    docker_networks:
+      - name: bbnet
+        ipam_config:
+          - subnet: '10.10.0.0/16'
```
btravouillon, line 20: Could you use 10.11.0.0/16 instead?

10.10.0.0/16 is widely used in the examples, I already implemented public (and private) environments with this subnet in [bluebanquise-vagrant](https://github.com/bluebanquise/bluebanquise-vagrant). Once the Vagrant env is up on a dev host, it is not possible to run this molecule test on the same dev host because the IP address will conflict between Vagrant and Docker.

osmocl: Corrected in commit : 879b3ae

Interesting the  [bluebanquise-vagrant](https://github.com/bluebanquise/bluebanquise-vagrant) project. Maybe not the place to discuss about it, but, why you do not use molecule ?
All the steps in this script shell [bootstrap.sh](https://github.com/bluebanquise/bluebanquise-vagrant/blob/master/tiny-hpc-cluster/bootstrap.sh) can be achieved with ansible tasks (in a `prepare.yml `file for example)
Also the [VagrantFile ](https://github.com/bluebanquise/bluebanquise-vagrant/blob/master/tiny-hpc-cluster/Vagrantfile) can be generated from a `molecule.yml` file

roles/core/dhcp_server/molecule/default/molecule.yml
```diff
@@ -17,11 +17,11 @@ platforms:
...
       - name: bbnet
         ipam_config:
-          - subnet: '10.10.0.0/16'
-            gateway: "10.10.0.254"
+          - subnet: '10.11.0.0/16'
+            gateway: "10.11.0.254"
     networks:
       - name: bbnet
-        ipv4_address: "10.10.0.1"            
+        ipv4_address: "10.11.0.1"
```
btravouillon, line 24:
```
[201] Trailing whitespace
molecule/default/molecule.yml:24
        ipv4_address: "10.11.0.1" 
```

.github/workflows/molecule-actions-core-dhcp_server.yml
btravouillon: On my workstation, the CentOS 8 test still fails while CentOS 7 succeeds (after a tweak in verify.yml to check dhcp package for EL7).

```
$ MOLECULE_DISTRO=centos:8 molecule test
[...]
    TASK [dhcp_server : package █ Install dhcp-server] *****************************
fatal: [management1]: FAILED! => {"changed": false, "msg": "Failed to download packages: Curl error (6):
Couldn't resolve host name for http://mirrorlist.centos.org/?release=8&arch=x86_64&repo=BaseOS&infra=container
[Could not resolve host: mirrorlist.centos.org]", "results": []}

ERROR: 
An error occurred during the test sequence action: 'converge'. Cleaning up.
```

```
$ MOLECULE_DISTRO=centos:7 molecule test
[...]
$ echo $?
0
```

The use of firewalld in Docker containers is not standard per se, it looks like it triggers some bad behaviour when we define additional Docker networks. I understand you defined this additional network to be able to start the dhcp service with the expected IP address. I had the same issue with dns_server and eventually test the role with `start_service: false`.

Would it be possible to remove the network bbnet from this scenario? It would allow to test the role with CentOS 7 and 8 (both are supported) with the firewall enabled. We can still consider another scenario to do advanced test, without the firewalld service.

Note: I have some code sleeping in my computer for months for a new scenario for dns_server (now advanced_dns_server) that uses Docker networks the way you do here. I will refresh it and push to a new branch. We might converge to some good practice.

btravouillon: BTW, you can test if the dhcpd.conf is correct by running `dhcpd -t -cf /etc/dhcp/dhcpd.conf` instead of starting the service. It can be a good test to add to the verify.yml.

oxedions: In a stack we made 2 years ago, there was some tests like that. It was pretty cool !

roles/core/dhcp_server/molecule/default/molecule.yml
btravouillon: To be honest I started this project one week before starting to play with molecule and did not think of it.

I'm using this project to quickly setup several environments with respect to BB's ansible inventory structure, for development and for testing. I may use it later for trainings as well, to quickly spawn environments for students. I need it to be compatible with libvirt and Virtualbox (and more recently Hyper-V which is still WIP), I don't know if molecule can create the relevant VagrantFile for all these backends.

If you want to discuss this further, feel free to open a new [discussion](https://github.com/bluebanquise/bluebanquise/discussions) 😀

.github/workflows/molecule-actions-core-dhcp_server.yml
osmocl: I prefere to keep the firewall tests as-is because we need to cover/test most of the tasks of this role.
We do not need to run `dhcpd -t -cf /etc/dhcp/dhcpd.conf` if the `dhcpd `service is correctly started.
In the past I already had several problems creating a new docker network, it always gave me a headache !
I believed that the easiest way to manage the docker network is using environment variables before running `molecule test`.
We can discover the container ip using `docker inspect.` So basically we should have something like this (a little dirty but it should work): 

```
# cat .github/workflows/molecule-actions-core-dhcp_server.yml

      [...]
	  
      - name: Run molecule test for core/dhcp_server
        run: |
	  export IP_CONTAINER=$(docker inspect -f "{{ .NetworkSettings.IPAddress }}" management1)
	  export IP_PREFIX_SUBNET1=$(docker inspect -f "{{ .NetworkSettings.IPAddress }}" management1 | cut -d "." -f 1-2)
          cd roles/core/dhcp_server && molecule test
```

Finally the provisioner section in `molecule.yml` will look like:

```
# cat molecule/default/molecule.yml

[...]

provisioner:
  name: ansible
  inventory:
    host_vars:
      management1:

        network_interfaces:
          - interface: enp0s3
            ip4: ${IP_CONTAINER}
            mac: 08:00:27:dc:f8:f5
            network: ice1-1

        [...]

        networks:
          ice1-1:
            subnet: ${IP_PREFIX_SUBNET1}.0.0
            prefix: 16
            netmask: 255.255.0.0
            broadcast: ${IP_PREFIX_SUBNET1}.255.255
            dhcp_unknown_range: ${IP_PREFIX_SUBNET1}.254.1 ${IP_PREFIX_SUBNET1}.254.254
            gateway: ${IP_PREFIX_SUBNET1}.2.1
            is_in_dhcp: true
            is_in_dns: true
            services_ip:
              pxe_ip: ${IP_CONTAINER}
              dns_ip: ${IP_CONTAINER}
              repository_ip: ${IP_CONTAINER}
              authentication_ip: ${IP_CONTAINER}
              time_ip: ${IP_CONTAINER}
              log_ip: ${IP_CONTAINER}

    hosts:
      compute:
        hosts:
          compute1:
            j2_current_iceberg_network: ice1
            network_interfaces:
            - interface: enp0s3
              ip4: ${IP_PREFIX_SUBNET1}.3.1
              mac: 08:00:27:dc:f8:f6
              network: ice1-1
```

This approach allow us to start the services `firewalld `and `dhcpd` using the default docker network.
What do you think about it?

btravouillon: > I prefere to keep the firewall tests as-is because we need to cover/test most of the tasks of this role.
> We do not need to run `dhcpd -t -cf /etc/dhcp/dhcpd.conf` if the `dhcpd `service is correctly started.

Yes, I also prefer to keep the firewall tests. The issue here seems to be between firewalld and the Docker network `bbnet`. The role needs this network for a single purpose: start the `dhcpd` service. If the role does not start the service but verify the configuration with `dhcpd -t -cf /etc/dhcp/dhcpd.conf`, we're all good.

> In the past I already had several problems creating a new docker network, it always gave me a headache !

Yeah, this was somewhat limited until molecule 3.0.5 release in July.

> I believed that the easiest way to manage the docker network is using environment variables before running `molecule test`.
> We can discover the container ip using `docker inspect.` So basically we should have something like this (a little dirty but it should work):
> [...]
> This approach allow us to start the services `firewalld `and `dhcpd` using the default docker network.
> What do you think about it?

Did you test this?
I believe the action will evaluate the docker inspect command before the start of molecule, when container management1 does not exist yet.

osmocl: It was only an idea but you are right the container does not exist yet, we need to create a dummy container before running `molecule test` , this container will be destroyed by the default scenario at the beginning of the molecule execution.
I am thinking in something like this, but I am not really convinced with this workaround: 

```
# cat .github/workflows/molecule-actions-core-dhcp_server.yml

      [...]
	  
      - name: Run molecule test for core/dhcp_server
        run: |
	  docker pull quay.io/actatux/ansible-centos:8
          docker create --name management1 quay.io/actatux/ansible-centos:8
          docker start management1
	  export IP_CONTAINER=$(docker inspect -f "{{ .NetworkSettings.IPAddress }}" management1)
	  export IP_PREFIX_SUBNET1=$(docker inspect -f "{{ .NetworkSettings.IPAddress }}" management1 | cut -d "." -f 1-2)
          cd roles/core/dhcp_server && molecule test
```
		  
I have tested it in my local environment and it works for centos7 and centos8.

For now I only see two solutions:

- either we test only in a centos8 container with the declarative docker network (my first proposition) 
- or we use the workaroud with environments vars to be able to test centos7 and centos8 containers (my second proposition) 

We need to decide soon which solution to use, actually I am stuck with the roles using networks (cf: dhcp, advanced-dhcp, pxe_stack, etc) due to this.

If you have another idea it will be more than welcome :)

btravouillon: > either we test only in a centos8 container with the declarative docker network (my first proposition)

No, it's not reliable. CentOS 8 does not work on my workstation, CentOS 7 does.

> or we use the workaroud with environments vars to be able to test centos7 and centos8 containers (my second proposition)

That's definitely too hacky. We will quickly run in troubles when spawning several containers.

> If you have another idea it will be more than welcome :)

If you really want to stick with testing firewalld + start services that need a dedicated Docker network in the same molecule scenario, my idea is to find a way to make it work out-of-the-box, as it does with default network. Running firewalld inside a container is not a common use-case since its backend is running in-kernel, it's not a surprise to have strange behaviour. That being said, we might miss a tuning somewhere?

Why do you need the `bbnet` Docker network in this scenario?

osmocl: > Why do you need the bbnet Docker network in this scenario?


In order to run the dhcpd service I need to know the subnet/Gateway and the IP to configure the dhcp server and a fake compute node, that is why I am declaring the `bbnet` Docker network.
Perhaps I am going a little too far and as you said a simple `dhcpd -t -cf /etc/dhcp/dhcpd.conf ` could be enough as the more advanced unitary test.

Additionally more advanced test (including functional tests) could be performed using virtual machines in our internal platform. I mean every roles that need a specific netwotk will be covered in this part.

btravouillon: The fake compute node is not a container, it does not need any Docker network. The ansible controller will load the inventory and create the dhcp configuration from the template.

Question is: Why do you need to run the dhcpd service?

If it's to test the configuration is correct, the `dhcpd -t ...` might be sufficient, no need to start the service.

This is the default scenario. We can then create more scenario without firewalld to add Docker networks and start some services. (The firewall configuration for the role being tested in the default scenario, we don't need to test it further).

osmocl: > Question is: Why do you need to run the dhcpd service?

Yes, I want to test if the configuration is correct via the service.
For now, I will rework this PR and the other one (advanced_dhcp_server) based on our discussion.
We will modify these tests in the future once we find a better solution concerning the docker network.
Thanks.

hmeAtos:
Hello,
I just remembered that the python module `python3-netaddr.noarch` will be required to be able to calculate the DHCP server configuration files.
Can we add this verification here?
Normally it's the `bluebanquise `role molecule test that does this verification. https://github.com/bluebanquise/bluebanquise/blob/be64686abdb5dfc763691f238b38823223a8b96c/roles/core/bluebanquise/molecule/default/verify.yml#L7
It is smart to re-check here?
Regards

btravouillon:
> I just remembered that the python module `python3-netaddr.noarch` will be required to be able to calculate the DHCP server configuration files.
> Can we add this verification here?

This module is required on the ansible controller, not the dhcp server. It is explicitly installed [here](https://github.com/bluebanquise/bluebanquise/pull/413/files#diff-84d28cb2165dbda4cdc76d46e504706d3b05108abe3995dbe7f933d7fa1fc69eR36) for this scenario, there is no need to verify it on the ansible controller.

hmeAtos:
> > I just remembered that the python module `python3-netaddr.noarch` will be required to be able to calculate the DHCP server configuration files.
> > Can we add this verification here?
> 
> This module is required on the ansible controller, not the dhcp server. It is explicitly installed [here](https://github.com/bluebanquise/bluebanquise/pull/413/files#diff-84d28cb2165dbda4cdc76d46e504706d3b05108abe3995dbe7f933d7fa1fc69eR36) for this scenario, there is no need to verify it on the ansible controller.

Great,
That's what I wanted to know, 
Thanks @btravouillon

btravouillon (approved):
Looks good to me. 👍
Thank you @osmocl.

oxedions (commented):
Dear @osmocl 
Could you also increment role version, like you did here for nfs_server:

https://github.com/bluebanquise/bluebanquise/pull/412/files

btravouillon:
> Could you also increment role version, like you did here for nfs_server:

Well, [I explicitly asked him to remove it](https://github.com/bluebanquise/bluebanquise/pull/413#discussion_r504810766). There is no functional change in the role, only a new molecule scenario.

oxedions:
> 
> 
> > Could you also increment role version, like you did here for nfs_server:
> 
> Well, [I explicitly asked him to remove it](https://github.com/bluebanquise/bluebanquise/pull/413#discussion_r504810766). There is no functional change in the role, only a new molecule scenario.

OK. Since it was incremented for the nfs_server, I was wondering. Let's keep it like this.

btravouillon:
> > > Could you also increment role version, like you did here for nfs_server:
> > 
> > 
> > Well, [I explicitly asked him to remove it](https://github.com/bluebanquise/bluebanquise/pull/413#discussion_r504810766). There is no functional change in the role, only a new molecule scenario.
> 
> OK. Since it was incremented for the nfs_server, I was wondering. Let's keep it like this.

That's a mistake. nfs_client introduced a new variable, not nfs_server.

PR #426, opened 2020-10-19
Feat: pxe_stack - Add fbtftp for RHEL/CentOS 8

oxedions:
For 8 major only.

Also still possible to force atftp installation using precedence.

### DIFF
--- a/roles/core/pxe_stack/readme.rst
+++ b/roles/core/pxe_stack/readme.rst
@@ -173,7 +173,7 @@ To be done
 Changelog
 ^^^^^^^^^
 
-* 1.1.6: Add ability to handle other tftp server than atftp. Benoit Leveugle <benoit.leveugle@gmail.com>
+* 1.1.6: Add ability to install other tftp server than atftp. Benoit Leveugle <benoit.leveugle@gmail.com>
 * 1.1.5: Update role to match $basearch, add status feat to bootset. Benoit Leveugle <benoit.leveugle@gmail.com>
 * 1.1.4: Update to new network_interfaces syntax. Benoit Leveugle <benoit.leveugle@gmail.com>
 * 1.1.3: Clean. johnnykeats <johnny.keats@outlook.com>
--- a/roles/core/pxe_stack/readme.rst
+++ b/roles/core/pxe_stack/readme.rst
@@ -153,6 +153,18 @@ All files are manually editable. Also, note that an unregistered host (so no hos
 
 To follow the deployment process, simply tail -f logs of http server, and see the whole process occurring.
 
+Optional parameters
+^^^^^^^^^^^^^^^^^^^
+
+It is possible to use another tftp server than atftp, using variables. By
+default, BlueBanquise relies on atftp, but fbtftp is also provided.
+
+To set the desired tftp server, and override default (atftp), define in the
+inventory or in the playbook the following variables:
+
+* **pxe_stack_tftp_package**: set the package name of the tftp server to be used. Stack propose *atftp* or *fbtftp_server*.
+* **pxe_stack_tftp_service**: set the service name of the tftp server to be used. Stack propose *atftpd* or *fbtftp_server*.
+
 To be done
 ^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


btravouillon:
Did you try to force install atftp with this change?

oxedions:
You mean using variables ? You have to redefine both list to make it work 🙁 

I didn't found a simple way to achieve the switch, without making it complex.

btravouillon:
I mean to effectively install atftp instead of fbtftp. We load the vars in the first task of the role, it will take precedence over any other value, doesn't it?

oxedions:
I was thinking to extra-vars as a test, but to make it permanent yes, that is an issue...

Then maybe I we can deprecate atftp for 1.4 and assume fbtftp is to be the default and only one.

If not, I need to modify too much the role.

btravouillon:
Can we consider this?

We force the default tftp package in vars (`pxe_stack_default_tftp`), but we allow to define `pxe_stack_tftp` in the inventory. If this variable exists, do not install the default tftp package.
I would recommand to apply the change to all vars/*.yml to ensure `pxe_stack_default_tftp` is defined.

It would allow to introduce fbtftp as tech-preview in 1.4: default will be atftp, but any user can opt-in to use fbtftp. Then in a later release, we can consider changing the default to fbtftp if feedbacks are good.

```
diff --git a/roles/core/pxe_stack/vars/RedHat_8.yml b/roles/core/pxe_stack/vars/RedHat_8.yml
index 42aa470..076204f 100644
--- a/roles/core/pxe_stack/vars/RedHat_8.yml
+++ b/roles/core/pxe_stack/vars/RedHat_8.yml
@@ -3,7 +3,6 @@ pxe_stack_apache_cgi_module: cgid
 pxe_stack_apache_conf_path: /etc/httpd/conf.d
 pxe_stack_apache_user: apache
 pxe_stack_packages_to_install:
-  - atftp
   - ipxe-x86_64-bluebanquise
   - ipxe-arm64-bluebanquise
   - httpd
@@ -11,6 +10,7 @@ pxe_stack_packages_to_install:
   - clustershell
   - python3-clustershell
   - policycoreutils
+pxe_stack_default_tftp: atftp
 pxe_stack_services_to_start:
   - atftpd
   - httpd
diff --git a/roles/core/pxe_stack/tasks/main.yml b/roles/core/pxe_stack/tasks/main.yml
index 91e0960..1f2ad10 100644
--- a/roles/core/pxe_stack/tasks/main.yml
+++ b/roles/core/pxe_stack/tasks/main.yml
@@ -35,9 +35,10 @@
   tags:
     - firewall
 
-- name: "package █ Install {{ pxe_stack_packages_to_install | join(' ') }} packages"
+- name: "package █ Install {{ (pxe_stack_packages_to_install +
+                [pxe_stack_tftp | default(pxe_stack_default_tftp)]) | join(' ') }} packages"
   package:
-    name: "{{ pxe_stack_packages_to_install }}"
+    name: "{{ pxe_stack_packages_to_install + [pxe_stack_tftp | default(pxe_stack_default_tftp)] }}"
     state: present
   tags:
     - package
```

oxedions:
Seems ok to me. Give me few hours to compute that.

btravouillon:
I can update the PR with this code if you want to.

oxedions:
Ok it is ok for me. I will update the code right now 😊

roles/core/pxe_stack/readme.rst
```diff
@@ -161,6 +161,7 @@ To be done
 Changelog
 ^^^^^^^^^
 
+* 1.1.6: Replace atftp by fbtftp by default on RHEL/CentOS 8. Benoit Leveugle <benoit.leveugle@gmail.com>
```
btravouillon, line 164: Needs an update.

btravouillon (changes requested):
Can you update the vars/*.yml files for other OS as well?
Otherwise the role will fail to default(pxe_stack_default_tftp_foo).

Can you also update the readme of the role to document the 2 new optional parameters?

roles/core/pxe_stack/readme.rst
oxedions: Yup. Should be ok now.

roles/core/pxe_stack/readme.rst
```diff
@@ -161,7 +173,7 @@ To be done
 Changelog
 ^^^^^^^^^
 
-* 1.1.6: Replace atftp by fbtftp by default on RHEL/CentOS 8. Benoit Leveugle <benoit.leveugle@gmail.com>
+* 1.1.6: Add ability to handle other tftp server than atftp. Benoit Leveugle <benoit.leveugle@gmail.com>
```
btravouillon, line 176: ...to install other... ? Or do we handle more than the installation?

oxedions: Haha, ok, updating.

btravouillon (approved):
Looks good to me. 👍

Issue #429, opened 2020-10-20
Missing packages slurm_exporter and ha_cluster_exporter in repositories


osmocl:
The package `slurm_exporter` and `ha_cluster_exporter` described in `roles/addons/prometheus_client/readme.rst` are missing in bluebanquise repositories: [https://bluebanquise.com/repository/releases/1.3/el8/x86_64/bluebanquise/](https://bluebanquise.com/repository/releases/1.3/el8/x86_64/bluebanquise/)

```
# cat roles/addons/prometheus_client/readme.rst
[...]

  monitoring:
    exporters:
      node_exporter:
        package: node_exporter
        service: node_exporter
        port: 9100
      ha_cluster_exporter:
        package: ha_cluster_exporter
        service: ha_cluster_exporter
        port: 9664
      slurm_exporter:
        package: slurm_exporter
        service: slurm_exporter
        scrape_interval: 5m
        scrape_timeout: 5m
        port: 9817
[...]
```

btravouillon:
The BlueBanquise project does not ship ha_exporter. We may remove this from the README example, while I find it interesting to demonstrate the role prometheus_client is expandable.

@oxedions: do you plan to provide slurm_exporter? If so, we might move this ticket to bluebanquise-infrastructure.

oxedions:
Hello.

The slurm_exporter is pretty complex to build. In the short term, I do not have the bandwidth to add it 🙁 
But Yes, I intend to release this one with the others.

The ha_exporter is indeed not in the scope of the stack, but can be easily added into an RPM as the team provides AMD64 binaries.

btravouillon:
Should we update the documentation to remove those examples? Or add a sentence to explain it's up to the user/admin to build and install these exporters? Or let it like that?

oxedions:
I would like to keep them in the main documentation. But yes, I will add sentences to explain that these are not provided by us.

oxedions:
Closing, since PR is done.

Issue #366, opened 2020-09-12
snmp_exporter fails to start


btravouillon:
## Description

The snmp_exporter.service fails to start.

```
# systemctl status snmp_exporter
● snmp_exporter.service - snmp_exporter
   Loaded: loaded (/etc/systemd/system/snmp_exporter.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Sat 2020-09-12 17:38:33 EDT; 1min 2s ago
  Process: 40244 ExecStart=/usr/local/bin/snmp_exporter --config.file=/etc/snmp_exporter/snmp_config.yml (code=exited, status=1/FAILURE)
 Main PID: 40244 (code=exited, status=1/FAILURE)

Sep 12 17:38:33 management1 systemd[1]: Started snmp_exporter.
Sep 12 17:38:33 management1 systemd[1]: snmp_exporter.service: Main process exited, code=exited, status=1/FAILURE
Sep 12 17:38:33 management1 systemd[1]: snmp_exporter.service: Failed with result 'exit-code'.
```

The journal reports an error parsing the config file:

```
Sep 12 17:38:33 management1 snmp_exporter[40244]: level=error ts=2020-09-12T21:38:33.192Z caller=main.go:156 msg="Error parsing config file" err="open /etc/snmp_exporter/snmp_config.yml: no such file or directory"
```

Indeed this file does not exist.
The role in charge of snmp_exporter installation should create the appropriate configuration file.

## Info

OS: CentOS 8.2
Branch: stable-1.3
snmp_exporter-0.18.0-1.el8.x86_64

oxedions:
This one is a bit tricky.
It must be generated using a dedicated tool: https://github.com/bluebanquise/bluebanquise/blob/master/resources/documentation/monitoring/snmp_exporter.rst

Maybe we could generate a "null" configuration file with the role, if and only it does not exist, then after that admin is in charge of the file (role will no more touch it if exist).
Other possibility, admin have to put a static file in "files" folder of the role if activation snmp.

btravouillon:
[Upstream project](https://github.com/prometheus/snmp_exporter) provides a default `snmp.yml` config file.
We can either:

1. Provide the upstream `snmp.yml` config file in the RPM (infra)
2. Provide our own `snmp.yml` config file in the RPM (infra)
[...]

PR #430, opened 2020-10-20
Adding unitary molecule tests for pormetheus_server/client roles

osmocl:
Molecule tests for:

- prometheus_server
- prometheus_client

Note that in `prometheus_server `role, the `snmp_exporter `service is not started and that cause an idempotency fail.
This error is tracked and being discussed here: [https://github.com/bluebanquise/bluebanquise/issues/366](https://github.com/bluebanquise/bluebanquise/issues/366)

Also packages `slurm_exporter` and `ha_cluster_exporter` described in `roles/addons/prometheus_client/readme.rst` are missing in bluebanquise repositories.
Issue open here: [https://github.com/bluebanquise/bluebanquise/issues/429](https://github.com/bluebanquise/bluebanquise/issues/429)

### DIFF
--- a/roles/addons/prometheus_client/molecule/default/molecule.yml
+++ b/roles/addons/prometheus_client/molecule/default/molecule.yml
@@ -13,6 +13,7 @@ platforms:
       - /sys/fs/cgroup:/sys/fs/cgroup:ro
     privileged: true
     pre_build_image: true
+
 provisioner:
   name: ansible
   inventory:
@@ -29,9 +30,5 @@ provisioner:
               package: node_exporter
               service: node_exporter
               port: 9100
-            ha_cluster_exporter:
-              package: ha_cluster_exporter
-              service: ha_cluster_exporter
-              port: 9664
 verifier:
   name: ansible
--- a/roles/addons/prometheus_client/molecule/default/molecule.yml
+++ b/roles/addons/prometheus_client/molecule/default/molecule.yml
@@ -1,0 +1,37 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+platforms:
+  - name: instance
+    image: "quay.io/actatux/ansible-${MOLECULE_DISTRO:-centos:8}"
+    override_command: false
+    capabilities:
+      - "SYS_ADMIN"
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
+    pre_build_image: true
+provisioner:
+  name: ansible
+  inventory:
+    host_vars:
+      instance:
+
+        start_services: true
+        enable_services: true
+        ep_firewall: true
+
+        monitoring:
+          exporters:
+            node_exporter:
+              package: node_exporter
+              service: node_exporter
+              port: 9100
+            ha_cluster_exporter:
+              package: ha_cluster_exporter
+              service: ha_cluster_exporter
+              port: 9664
+verifier:
+  name: ansible
--- a/roles/addons/prometheus_client/molecule/default/prepare.yml
+++ b/roles/addons/prometheus_client/molecule/default/prepare.yml
@@ -9,7 +9,6 @@
       when: ansible_facts.os_family == "RedHat"
       loop:
         - firewalld
-        - httpd
 
     - name: Start needed services
       service:
@@ -19,7 +18,6 @@
       when: ansible_facts.os_family == "RedHat"
       loop:
         - firewalld
-        - httpd
 
     - name: "Configure BlueBanquise repository"
       yum_repository:
--- a/roles/addons/prometheus_client/molecule/default/prepare.yml
+++ b/roles/addons/prometheus_client/molecule/default/prepare.yml
@@ -1,0 +1,30 @@
+---
+- name: Prepare
+  hosts: all
+  tasks:
+    - name: Install needed packages
+      package:
+        name: "{{ item }}"
+        state: present
+      when: ansible_facts.os_family == "RedHat"
+      loop:
+        - firewalld
+        - httpd
+
+    - name: Start needed services
+      service:
+        name: "{{ item }}"
+        enabled: yes
+        state: started
+      when: ansible_facts.os_family == "RedHat"
+      loop:
+        - firewalld
+        - httpd
+
+    - name: "Configure BlueBanquise repository"
+      yum_repository:
+        name: bluebanquise
+        description: BlueBanquise 1.3 repository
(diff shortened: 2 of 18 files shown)


roles/addons/prometheus_server/molecule/default/molecule.yml
```diff
@@ -0,0 +1,57 @@
...
+    test_sequence:
+    - dependency
+    - lint
+    - cleanup
+    - destroy
+    - syntax
+    - create
+    - prepare
+    - converge
+    #- idempotence
```
btravouillon, line 53: Can you add a comment `# FIXME idempotence disabled because #366` above?

I wonder if we could only skip the idempotence for this specific task? You may have to modify the role itself and add the tag `molecule-idempotence-notest` to the task. (+ the comment above on why we skip it)
It would allow to run idempotence test for all other tasks, which is IMHO a better option.

roles/addons/prometheus_server/molecule/default/prepare.yml
```diff
@@ -0,0 +1,30 @@
...
+  hosts: all
+  tasks:
+    - name: Install needed packages
+      package:
+        name: "{{ item }}"
+        state: present
+      when: ansible_facts.os_family == "RedHat"
+      loop:
+        - firewalld
+        - httpd
```
btravouillon, line 12: Why do you need to install httpd during the verify step?

roles/addons/prometheus_server/molecule/default/prepare.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: "Configure BlueBanquise repository"
+      yum_repository:
+        name: bluebanquise
+        description: BlueBanquise 1.3 repository
+        baseurl: https://bluebanquise.com/repository/releases/1.3/el$releasever/$basearch/bluebanquise/
```
btravouillon, line 28: You can use latest instead of 1.3

btravouillon (changes requested):
Thank you for this new test. Some comments in line. Can you also take a look at the failure/warning from the static code analysis?

roles/addons/prometheus_client/molecule/default/prepare.yml
```diff
@@ -0,0 +1,30 @@
...
+  hosts: all
+  tasks:
+    - name: Install needed packages
+      package:
+        name: "{{ item }}"
+        state: present
+      when: ansible_facts.os_family == "RedHat"
+      loop:
+        - firewalld
+        - httpd
```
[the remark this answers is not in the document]
osmocl: no need 😃

roles/addons/prometheus_client/molecule/default/verify.yml
```diff
@@ -0,0 +1,59 @@
...
+      assert:
+        that: firewall_cmd_result.stdout |
+                    regex_findall(('9664/tcp'), multiline=False) | length == 1
+
+    - name: Assert node_exporter and ha_cluster_exporter packages are installed
+      assert:
+        that: "'{{item}}' in ansible_facts.packages"
+      loop:
+        - node_exporter
+        - ha_cluster_exporter
```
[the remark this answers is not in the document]
osmocl: removed, already disscused in #429

roles/addons/prometheus_client/molecule/default/molecule.yml
```diff
@@ -0,0 +1,37 @@
...
+        enable_services: true
+        ep_firewall: true
+
+        monitoring:
+          exporters:
+            node_exporter:
+              package: node_exporter
+              service: node_exporter
+              port: 9100
+            ha_cluster_exporter:
```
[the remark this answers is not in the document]
osmocl: removed, already disscused in #429

roles/addons/prometheus_server/molecule/default/molecule.yml
osmocl: I prefer to keep the line commented until #366 is fixed.
The prometheus roles should be reworked.

roles/addons/prometheus_server/molecule/default/prepare.yml
osmocl: no need

roles/addons/prometheus_server/molecule/default/molecule.yml
btravouillon: Can you consider the change below instead?

```
diff --git a/roles/addons/prometheus_server/molecule/default/molecule.yml b/roles/addons/prometheus_server/molecule/default/molecule.yml
index 8b2c23b..c4ab9f7 100644
--- a/roles/addons/prometheus_server/molecule/default/molecule.yml
+++ b/roles/addons/prometheus_server/molecule/default/molecule.yml
@@ -40,18 +40,3 @@ provisioner:

 verifier:
   name: ansible
-scenario:
-    test_sequence:
-    - dependency
-    - lint
-    - cleanup
-    - destroy
-    - syntax
-    - create
-    - prepare
-    - converge
-    #- idempotence # FIXME idempotence disabled because #366
-    - side_effect
-    - verify
-    - cleanup
-    - destroy
diff --git a/roles/addons/prometheus_server/tasks/main.yml b/roles/addons/prometheus_server/tasks/main.yml
index 3db4116..d392ebe 100644
--- a/roles/addons/prometheus_server/tasks/main.yml
+++ b/roles/addons/prometheus_server/tasks/main.yml
@@ -157,3 +157,4 @@
   when: (start_services | bool)
   tags:
     - service
+    - molecule-idempotence-notest # FIXME idempotence disabled because #366
```

This skips idempotence test on the failing task only. All other tasks will be tested, which provides a better coverage.

osmocl: Nice, I did not know the tag `molecule-idempotence-notest`
I will check it

osmocl:
While running unitary tests in centos7, I found out that the prometheus service is not being added to the firewall.
See error here: https://github.com/bluebanquise/bluebanquise/pull/430/checks?check_run_id=1363786120#step:5:1004
	
To solve this issue, I added in my last commit the following firewall configuration for prometheus service :

```
# cat roles/addons/prometheus_server/files/prometheus.xml
<?xml version="1.0" encoding="utf-8"?>
<service>
  <short>prometheus</short>
  <description>Prometheus server</description>
  <port protocol="tcp" port="9090"/>
</service>
```

oxedions:
Dear @osmocl 

Would you be able to move this PR on the community repository ?

PR #431, opened 2020-10-20
Fix: core/nic - Update readme

oxedions:
Fix readme, since having slave NIC first in the list break the rules defined in the CHANGELOG.md:

1. The main resolution network of hosts is the value of the network parameter of the first item of the list (e.g., c001 will be on the same line than c001-ice1-1 in the hosts file).
2. First management network related item in the list will be the ansible main ssh target interface (from ssh_master role), and also the main management network interface for the client (services_ip to target on client side).

### DIFF
--- a/roles/core/nic/readme.rst
+++ b/roles/core/nic/readme.rst
@@ -38,18 +38,18 @@ To configure an LACP bonding, specify slave interfaces, and then create the bond
 .. code-block:: yaml
 
   network_interfaces:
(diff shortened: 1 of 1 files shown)


btravouillon (approved):
Looks good to me. 👍
Nice catch!

oxedions:
Well, its not me, a colleague reported an issue with the hosts_file role, and that was the source 🙂

PR #434, opened 2020-10-20
fix: execute the gather variables task even with specific tags

btravouillon:
For each role, the task "include_vars ░ Gather OS specific variables" **MUST** be executed whatever the selection of tags.

Otherwise, other tasks in the role will fail because required variables are not defined.

Example (before this change):
```
# ansible-playbook playbooks/managements.yml -l management1 --tags firewall
[WARNING]: Unable to parse /etc/bluebanquise/internal as an inventory source

PLAY [managements playbook] ******************************************************

TASK [Gathering Facts] ***************************************************************
Tuesday 20 October 2020  19:22:09 -0400 (0:00:00.067)       0:00:00.067 *******
ok: [management1]

PLAY [managements playbook] ******************************************************
TASK [Gathering Facts] ***************************************************************
Tuesday 20 October 2020  19:22:11 -0400 (0:00:02.622)       0:00:02.689 *******
ok: [management1]

TASK [repositories_server : firewalld █ "Add services to firewall's public zone"] *
Tuesday 20 October 2020  19:22:13 -0400 (0:00:01.753)       0:00:04.443 *******
fatal: [management1]: FAILED! => 
  msg: '''repositories_server_firewall_services_to_add'' is undefined'

PLAY RECAP **************************************************************************
management1 : ok=2 changed=0 unreachable=0 failed=1 skipped=0 rescued=0 ignored=0
```

Indeed, `repositories_server_firewall_services_to_add` is not defined since the role did not gather the variables.

I use this PR to add coherency between roles: use same task name, same tags, etc.


oxedions:
Dear @btravouillon 

Thank you for this PR 🙂 
You are right, this is an issue that must be solved.

I think in my side that I would prefer to add tags these include are related of.

For example in a role containing packages, services, and firewall related tasks (tagged) then I would add tags package, service and firewall.

Having always for each of these includes may result in some kind of overhead when executing only a single role (which happen much more time on my side than executing the whole playbook or a specific task tag).

bouriquet:
> 
> 
> Dear @btravouillon
> 
> Thank you for this PR 🙂
> You are right, this is an issue that must be solved.
> 
> I think in my side that I would prefer to add tags these include are related of.
> 
> For example in a role containing packages, services, and firewall related tasks (tagged) then I would add tags package, service and firewall.
> 
> Having always for each of these includes may result in some kind of overhead when executing only a single role (which happen much more time on my side than executing the whole playbook or a specific task tag).

Agree with this solution

hmeAtos (approved):
ok for me @btravouillon 
thanks

oxedions:
@btravouillon Do you want me to update the PR with this solution ?

btravouillon:
No I will.

btravouillon:
I wrote a quick & dirty helper script yesterday evening to ease the maintenance of these tags.

For each role, the logic is:

1. Search for any variable in `vars/*.yml`
2. For each variable, search any usage in templates or tasks
3. If found, collect the tags of the relevant tasks (including the template task if any)
4. Compare the tags with the list of tags of the first task "include_vars ░ Gather OS specific variables"
5. Report an error if some tags are missing for include_vars.

I added this script to the static code analysis in the CI, current result below: https://github.com/bluebanquise/bluebanquise/pull/434/checks?check_run_id=1295868755

I may have forgotten some use cases, feel free to comment.

roles/core/bluebanquise/tasks/main.yml
```diff
@@ -20,6 +20,7 @@
       skip: true
   tags:
     - internal
+    - always
```
btravouillon, line 23: not detected

roles/core/dns_client/tasks/main.yml
```diff
@@ -20,6 +20,7 @@
       skip: true
   tags:
     - internal
+    - always
```
btravouillon, line 23: not detected

btravouillon:
BTW, can we remove (or rename) the file [roles/core/pxe_stack/tasks/Ubuntu_18.yml](https://github.com/bluebanquise/bluebanquise/blob/master/roles/core/pxe_stack/tasks/Ubuntu_18.yml)?

oxedions:
You can remove it. Ubuntu 18.04 should not be supported any more, and the 20.04 is a mess to deploy, so no way currently to use it.

roles/core/bluebanquise/tasks/main.yml
bouriquet: why `always ` instead of `package ` tag ?

btravouillon: The helper script gather_vars_checker.py did not detect the correct tags to use. I made a mistake in the script, I'm working on the fix at this exact moment :grin: .

bouriquet (approved):

oxedions:
Dear @btravouillon 

Thanks a lot for this PR. I will need few days to review it entirely with the python code. 🙂

oxedions (approved):
Dear @btravouillon 

Thank you for this PR.
Ok for me, merging.

PR #436, opened 2020-10-21
Fix: addons/report - update report role

oxedions:
Update to new inventory syntax.

Missing many patterns, but this takes so much time...


oxedions:
This should be ok for now. If needed, we can add more patterns later.

roles/addons/report/tasks/main.yml
```diff
@@ -81,10 +61,8 @@
   tags:
     - template
   when:
-    - report_settings.inventory_checker is defined
-    - report_settings.inventory_checker is not none
-    - report_settings.inventory_checker
-    - report_settings.inventory_checker is iterable
+    - report_inventory_checker is defined
```
btravouillon, line 64: It's in the defaults now, it is always defined.

roles/addons/report/tasks/main.yml
```diff
@@ -96,11 +74,9 @@
   tags:
     - template
   when:
-    - report_settings.inventory_checker is defined
-    - report_settings.inventory_checker is not none
-    - report_settings.inventory_checker
-    - report_settings.inventory_checker is iterable
-    - "'groups' in report_settings.inventory_checker"
+    - report_inventory_checker is defined
```
btravouillon, line 77: It's in the defaults now, it is always defined.

roles/addons/report/tasks/main.yml
```diff
@@ -112,11 +88,9 @@
   tags:
     - template
   when:
-    - report_settings.inventory_checker is defined
-    - report_settings.inventory_checker is not none
-    - report_settings.inventory_checker
-    - report_settings.inventory_checker is iterable
-    - "'hosts' in report_settings.inventory_checker"
+    - report_inventory_checker is defined
```
btravouillon, line 91: It's in the defaults now, it is always defined.

roles/addons/report/templates/graph-tool.py.j2
```diff
@@ -15,13 +15,13 @@ edge_color = g.new_edge_property("vector<float>")
 edge_text = g.new_edge_property("string")
 
 {% for network in networks %}
-{% if network not in report_settings.graph_tool_settings.excluded_networks %} 
+{#{% if network not in report_graph_tool_settings.excluded_networks %} #}
```
btravouillon, line 18: why the comment? was it a test?

roles/addons/report/templates/hosts.css.j2
```diff
@@ -24,12 +24,15 @@ padding: 10px 10px;
 }
 
 {% for host in groups[rack] %}
-
+{% if hostvars[host]['x_p'] is defined and hostvars[host]['z_p'] is defined and hostvars[host]['x_span'] is defined and hostvars[host]['z_span'] is defined %}
```
btravouillon, line 27: could you add some indentation for better readability? (will need the header `#jinja2: lstrip_blocks: True`)

roles/addons/report/templates/inventory_checker.html.j2
```diff
@@ -60,7 +60,7 @@ function includeHTML() {
 <br>
 Please consider that this debug tools will not catch all errors in your inventories, but try to provide protection again common mistakes.<br>
 <br>
-This report has been generated at:<br>
+This report has been generated at: {{ ansible_date_time.date}} {{ ansible_date_time.time }} {{ ansible_date_time.tz }}<br>
 {{ansible_managed}}<br>
```
btravouillon, line 64: Should you remove the `{{ansible_managed}}` ?

roles/addons/report/templates/inventory_checker_hosts.html.j2
```diff
@@ -21,13 +21,15 @@ Also check the main network exists and that all linked network exists #}
...
-      {% if hostvars[host]['network_interfaces'][interface]['network'] not in networks %}
-        <br><div id="error">ERROR</div>: host {{host}} interface {{interface}} has a network ({{hostvars[host]['network_interfaces'][interface]['network']}}) not defined in the configuration, please check.
+    {% if interface.network is defined and interface.network is not none %}
+      {% if interface.network not in networks %}
+        <br><div id="error">ERROR</div>: host {{ host }} interface {{ interface }} has a network ({{ interface.network }}) not defined in the configuration, please check.
       {% endif %}
     {% endif %}
-    {% if hostvars[host]['network_interfaces'][interface]['ip4'] is not defined or hostvars[host]['network_interfaces'][interface]['ip4'] is none %}
-        <br><div id="error">ERROR</div>: host {{host}} interface {{interface}} has no ip4 value. If you do not need this interface, please comment/erase it.
+    {% if interface.ip4 is not defined or interface.ip4 is none %}
```
btravouillon, line 29: You may combine both condition in the same if block:
`(interface.ip4 is not defined or interface.ip4 is none) and (interface.type is not defined or interface.type == 'ethernet')`

oxedions:
Dear @btravouillon 

Thanks a lot for these feedback. I will have a look :-)

roles/addons/report/tasks/main.yml
oxedions: Yes, I will remove that. Thanks for pointing that out.

roles/addons/report/templates/graph-tool.py.j2
oxedions: It was working in the first version and I want to keep a track of this as it lay be relevant later. But for now, I do not use it.

roles/addons/report/templates/hosts.css.j2
oxedions: sure !

roles/addons/report/templates/inventory_checker.html.j2
oxedions: yes, good catch, forgot that one.

roles/addons/report/templates/inventory_checker_hosts.html.j2
oxedions: Ok, I will investigate that part. Indeed, I can do simpler 🙂

roles/addons/report/templates/graph-tool.py.j2
btravouillon: Well, it is tracked in git history 😁 . If you want to keep it as comment in the code, should you add a [TODO or FIXME tag](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags)?

oxedions:
Hello,

I am closing this PR. After deep look, nobody use this role (except me), and maintaining it takes a huge amount of time.
I am thinking of a way to re-use all of this code in another format.

Thank you for the time made on reviewing this PR.

PR #444, opened 2020-10-23
Feat: addons/nic_nmcli - Add all options

oxedions:
Provide all features to nmcli module, while trying to keep retro compatibility.

Introducing new things for inventory, to be discussed. Again, trying to keep retro compatibility.

Also, discussion related to routing here (https://github.com/bluebanquise/bluebanquise/discussions/443) linked to this PR.

This role should replace core/nic once ready.


oxedions:
Hum. It seems Ansible team just updated the doc, and nmcli is now called community.general.nmcli. I should update it.
https://docs.ansible.com/ansible/latest/collections/community/general/nmcli_module.html

oxedions:
Forgot to set version in vars.

roles/addons/nic_nmcli/readme.rst
```diff
@@ -6,27 +6,106 @@ Description
...
+"""""""""""""""""""
+
+In multiple ip modes, you need to set the prefix yourself:
+
+.. code-block:: yaml
 
-Add more features, as nmcli Ansible module can do much more.
+  network_interfaces:
+    - interface: eth0
+      ip4: 10.10.0.1/16,10.10.0.2/16
```
bouriquet, line 61: could be great to have this as a list instead of coma separated string

roles/addons/nic_nmcli/tasks/main.yml
```diff
@@ -41,15 +41,66 @@
...
+    #     {{ item.ip4 }}
+    #   {% else %}\
+    #     {{ item.ip4 }}/{{ networks[item.network].prefix }}
+    #   {% endif %}
+    # {% else %}
+    #    omit
+    # {% endif %}
+    # Allows to provide a simple ip4, or an ip4/prefix
+    # Also support old prefix format (to be replaced by prefix4)
+    ip6: "{{ (item.ip6 is defined) | ternary( ('::' in (item.ip6|default('None')) ) | ternary( (item.ip6|default('None')), ( (item.ip6|default('None')) | string) + '::' + ((networks[item.network].prefix6|default('None')) |string)), omit) }}"
```
bouriquet, line 62: Is ipv6 fully supported by the stack ?

roles/addons/nic_nmcli/readme.rst
oxedions: I will try to do that 🙂

roles/addons/nic_nmcli/tasks/main.yml
oxedions: Not really. I was thinking of possible login nodes that would have 2 network, one internal in ipv4 and one external in ipv6.
Not sure it can really happen in fact...

oxedions: I will keep ipv6, but only as basic option, for external interfaces. Which means I will remove all 4/6 related elements. Should be done soon.

roles/addons/nic_nmcli/readme.rst
oxedions: Difficult honestly 🙁 We can do that in another PR

bouriquet (approved):

PR #451, opened 2020-10-27
ci: test timezone is 'Europe/Paris'

btravouillon:
Previous test was unreliable: when summer time is over, tz = CET.

Because we define the timezone to be 'Europe/Paris', check the
host configuration is updated to reflect that change. For this, check
/etc/localtime is a symlink to /usr/share/zoneinfo/Europe/Paris.

Tests are failing since Sunday when the time changed in Europe/Paris: https://github.com/bluebanquise/bluebanquise/runs/1310835559

### DIFF
--- a/roles/core/time/molecule/default/verify.yml
+++ b/roles/core/time/molecule/default/verify.yml
@@ -13,9 +13,14 @@
       debug:
         msg: "Timestamp format: {{ ansible_facts.date_time }}"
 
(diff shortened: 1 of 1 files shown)


oxedions:
Good catch !

This is in fact a fun bug 😄 

Ok for me, merging !

hmeAtos:
Thanks @btravouillon

PR #453, opened 2020-10-29
Feat: addons/system_state - role creation

oxedions:
System state role, to provide a last play state.


btravouillon:
I don't understand this role. Would the use of [ARA](https://ara.recordsansible.org/) answer the need?

oxedions:
Yes. This is why i stopped developing this PR and put it in draft, to investigate external tools like ARA deeper.

oxedions:
I am closing this PR. External tools seems to cover the need.

PR #454, opened 2020-10-29
Oxedions/quality

oxedions:
This is the last PR for this marathon.

* Cleaned old macros
* Added grafana and singularity to doc
* Converted singularity md to rst
* Updated/added all role versions

Starting from here, every modification to a roll MUST respect these:

* Update role version (even if changed one character)
* Ensure doc/readme are in rst format


hmeAtos (approved):

btravouillon (approved):
Looks good to me. 👍

btravouillon:
> * Update role version (even if changed one character)

I would not update the role version for non functional change. Any change to the CI (i.e. molecule scenario) won't change anything for the end-user, thus I don't see any reason to change the version.

oxedions:
Ok, I will create another PR for the CONTRIBUTE :-)

Thanks a lot for the review.

PR #461, opened 2020-11-03
Fix: core/display_tuning - Remove dynamic elements

oxedions:
Removing all dynamic elements.

Reason: this prevent scrolling, which is pretty annoying. I did not see that part when doing this role.

### DIFF
--- a/roles/core/display_tuning/files/screenrc
+++ b/roles/core/display_tuning/files/screenrc
@@ -4,8 +4,6 @@
 # - ProWi
 # - leo2501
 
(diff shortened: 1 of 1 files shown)


btravouillon (approved):
screenrc is rather personal. In the long run, I would make it a template, not a file. Anyway, looks good to merge as is.

oxedions:
> 
> 
> screenrc is rather personal. In the long run, I would make it a template, not a file. Anyway, looks good to merge as is.

I agree with you. I tried to propose a basic usable screenrc, but it seems indeed everyone has already its own.

PR #463, opened 2020-11-05
Conman : Improve molecule tests 

osmocl:
Adding molecule tests for conman role.

### DIFF
--- a/roles/core/conman/molecule/default/prepare.yml
+++ b/roles/core/conman/molecule/default/prepare.yml
@@ -1,6 +1,6 @@
 ---
 - name: Prepare
-  hosts: management1
+  hosts: all
 
   tasks:
     - name: "Configure BlueBanquise repository"
--- a/roles/core/conman/molecule/default/prepare.yml
+++ b/roles/core/conman/molecule/default/prepare.yml
@@ -12,3 +12,10 @@
       when:
         - ansible_facts.os_family == "RedHat"
         - ansible_facts.distribution_major_version == "8"
+
+    - name: Install needed package
+      package:
+        name: python2-rpm
+        state: present
+      when:
+        - ansible_facts.os_family == "Suse"
--- a/roles/core/conman/molecule/default/prepare.yml
+++ b/roles/core/conman/molecule/default/prepare.yml
@@ -6,7 +6,7 @@
     - name: "Configure BlueBanquise repository"
       yum_repository:
         name: bluebanquise
-        description: BlueBanquise 1.3 repository
-        baseurl: https://bluebanquise.com/repository/releases/1.3/el$releasever/$basearch/bluebanquise/
+        description: BlueBanquise repository
+        baseurl: https://bluebanquise.com/repository/releases/latest/el$releasever/$basearch/bluebanquise/
         gpgcheck: False
       when: ansible_facts.os_family == "RedHat"
--- a/roles/core/conman/molecule/default/prepare.yml
+++ b/roles/core/conman/molecule/default/prepare.yml
@@ -1,13 +1,12 @@
 ---
 - name: Prepare
-  hosts: all
+  hosts: management1
 
   tasks:
     - name: "Configure BlueBanquise repository"
       yum_repository:
         name: bluebanquise
-        description: BlueBanquise 1.2 repository
-        baseurl: https://bluebanquise.com/repository/releases/1.2/el$releasever/$basearch/bluebanquise/
+        description: BlueBanquise 1.3 repository
+        baseurl: https://bluebanquise.com/repository/releases/1.3/el$releasever/$basearch/bluebanquise/
         gpgcheck: False
       when: ansible_facts.os_family == "RedHat"
-
--- a/roles/core/conman/molecule/default/prepare.yml
+++ b/roles/core/conman/molecule/default/prepare.yml
@@ -9,4 +9,6 @@
         description: BlueBanquise repository
         baseurl: https://bluebanquise.com/repository/releases/latest/el$releasever/$basearch/bluebanquise/
         gpgcheck: False
-      when: ansible_facts.os_family == "RedHat"
+      when:
+        - ansible_facts.os_family == "RedHat"
+        - ansible_facts.distribution_major_version == "8"
--- a/roles/core/conman/molecule/default/verify.yml
+++ b/roles/core/conman/molecule/default/verify.yml
@@ -3,9 +3,18 @@
   hosts: all
   tasks:
 
-    - name: Collect package facts
+    - name: Collect package facts for RedHat
       package_facts:
         manager: auto
+      when:
+        - ansible_facts.os_family == "RedHat"
+
+    - name: Collect package facts for OpenSuse
+      command: rpm -qa | grep "conman|\ipmitool"
+      register: result
+      changed_when: false
+      when:
+        - ansible_facts.os_family == "Suse"
 
     - name: Collect services facts
       service_facts:
@@ -38,6 +47,13 @@
       when:
         - ansible_facts.os_family == "RedHat"
         - ansible_facts.distribution_major_version == "8"
+
+    - name: Assert needed OpenSuse conman packages are installed
+      assert:
+        that: result.stdout |
+                 regex_findall(('conman|ipmitool'), multiline=True) | length == 2
+      when:
+        - ansible_facts.os_family == "Suse"
 
     - name: Retrieve file /etc/conman.conf system status
       stat:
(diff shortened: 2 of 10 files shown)


btravouillon:
@oxedions Did you build conman for el7 on purpose? Should it replace the package from the OS? I would rather prefer to keep the conman from upstream.

oxedions:
@btravouillon No, it is a mistake. I did it too fast. Conman should not be provided on BB repositories on EL7. I will remove it.

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,101 @@
+---
+- name: Verify
+  hosts: management1
```
btravouillon, line 3: Prefer 'all' unless this is specific to a single host out of a group.

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,101 @@
...
+      service_facts:
+
+    - name: Assert common packages are installed
+      assert:
+        that: "'{{ item }}' in ansible_facts.packages"
+      loop:
+        - conman
+        - ipmitool
+      when:
+        - ansible_facts.os_family == "RedHat"
```
btravouillon, line 20: Check for OpenSUSE as well.

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,101 @@
...
+        that:
+          - reg_file_conman.stat.exists
+          - reg_file_conman.stat.mode == '0644'
+          - reg_file_conman.stat.pw_name == 'root'
+          - reg_file_conman.stat.gr_name == 'root'
+
+    - name: Retrieve client configuration in /etc/conman.conf file
+      lineinfile:
+        path: /etc/conman.conf
+        regexp: 'console name="compute1" dev="ipmitool.exp 10.10.102.1 ADMIN ADMIN"'
```
btravouillon, line 59: There is no need to create a fake host, you can check the line below:
`console name="instance" dev="ipmitool.exp 10.10.102.1 ADMIN ADMIN"`

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,101 @@
...
+        state: absent
+      check_mode: yes
+      register: reg_conf_conman
+      changed_when: false
+
+    - name: Check /etc/conman.conf file contains console configuration for expected client
+      assert:
+        that: reg_conf_conman.found
+
+    - name: Retrieve file /var/log/conman/compute1.log system status
```
btravouillon, line 69: Check: /var/log/conman/instance.log

btravouillon (changes requested):
Can you revert all your changes but the verify.yml and update of the baseurl?

roles/core/conman/molecule/default/prepare.yml
```diff
@@ -1,13 +1,12 @@
...
+  hosts: management1
 
   tasks:
     - name: "Configure BlueBanquise repository"
       yum_repository:
         name: bluebanquise
-        description: BlueBanquise 1.2 repository
-        baseurl: https://bluebanquise.com/repository/releases/1.2/el$releasever/$basearch/bluebanquise/
+        description: BlueBanquise 1.3 repository
+        baseurl: https://bluebanquise.com/repository/releases/1.3/el$releasever/$basearch/bluebanquise/
```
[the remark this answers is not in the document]
osmocl: Corrected in commit: 29794dc9d78147d2f7747418a84b6bdffa3ae419

roles/core/conman/molecule/default/verify.yml
osmocl: This test suite is executed only against the management.
As a compute is declared in `molecule.yml` if we declare `hosts: all` the tests will try to be executed also against the compute and the tests will fail.

roles/core/conman/molecule/default/verify.yml
osmocl: I wasn't able to run properly the role in opensuse 15.1.
Declaring an inventory variable `ansible_python_interpreter: /usr/bin/python3` does not solve this problem either.

ERROR:

```
TASK [Collect package facts] ***************************************************
Warning: 35m[WARNING]: Found "rpm" but Failed to import the required Python library (rpm)
on management1's Python /usr/bin/python. Please read the module documentation
and install it in the appropriate location. If the required library is
installed, but Ansible is using the wrong Python interpreter, please consult
the documentation on ansible_python_interpreter
fatal: [management1]: FAILED! => {"changed": false, "msg": "Could not detect a supported package manager from the following list: ['portage', 'pacman', 'rpm', 'pkg', 'apt'], or the required Python library is not installed. Check warnings for details."}
    
    PLAY RECAP *********************************************************************
    management1                : ok=1    changed=0    unreachable=0    failed=1
```

roles/core/conman/molecule/default/verify.yml
osmocl: I just added a fake compute host to be closer of a real installation/configuration.
IMHO, having a client configuration is helpful to read the molecule test suite and to understand it better.

roles/core/conman/molecule/default/verify.yml
btravouillon: Please don't change molecule.yml, instance is fine for the test.

roles/core/conman/molecule/default/verify.yml
btravouillon: I can't say, but last time this test ran against master, it was successful with OpenSUSE: https://github.com/bluebanquise/bluebanquise/actions/runs/329790788

Given that support of OpenSUSE is planned for the stack, I am not keen to remove any existing test.

roles/core/conman/molecule/default/verify.yml
btravouillon: There is no such thing as a real installation/configuration. Some will use baremetal management nodes, other will use containers or VMs.
The purpose of this unit test it to test the conman role. The role needs a host in the inventory to create a conman.conf file. `instance` is an host, and it is added to the configuration file during the run of the play. We don't need more than that today.

osmocl: Blah... I will modify this for you. Thanks.

roles/core/conman/molecule/default/verify.yml
osmocl: Tests added in latests commits

osmocl:
@oxedions, as discussed please do not forget to remove this package : 
https://bluebanquise.com/repository/releases/latest/el7/x86_64/bluebanquise/Packages/x86_64/conman-0.3.0-1.el7.x86_64.rpmhttps://bluebanquise.com/repository/releases/1.3/el7/x86_64/bluebanquise/Packages/x86_64/conman-0.3.0-1.el7.x86_64.rpm

Thanks in advance !

hmeAtos (dismissed):
ok for me
thank you very much @osmocl 
Regards

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,124 @@
+---
+- name: Verify
+  hosts: all
+  tasks:
+
+    - name: Collect package facts for RedHat
+      package_facts:
+        manager: auto
+      when:
+        - ansible_facts.os_family == "RedHat"
```
btravouillon, line 10: For OpenSUSE, the collect currently fails because the package `python2-rpm` is missing from the container. I will update the image whenever possible. In the meantime, a workaround would be to install `python2-rpm` in prepare.yml for OpenSUSE OS family.
This would allow to collect the packages for OpenSUSE as well, and ensure the packages are installed.

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,124 @@
...
+        path: /etc/conman.conf
+      register: reg_log_conman
+      changed_when: false
+
+    - name: Assert log client file /var/log/conman/instance.log exists
+      assert:
+        that: reg_log_conman.stat.exists
+
+
+    - name: Retrieve dir system status
```
btravouillon, line 78: Can you add the path to the name?

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,124 @@
...
+    - name: Assert dir conman_execpath /usr/share/conman/exec exist
+      assert:
+        that:
+          - reg_dir_conman.stat.exists
+          - reg_dir_conman.stat.isdir
+      when:
+        - ansible_facts.os_family == "RedHat"
+        - ansible_facts.distribution_major_version == "7"
+
+    - name: Retrieve dir system status
```
btravouillon, line 96: Can you add the path to the name?

roles/core/conman/molecule/default/verify.yml
```diff
@@ -0,0 +1,124 @@
...
+        - ansible_facts.distribution_major_version == "8"
+
+    - name: Assert dir conman_execpath /usr/lib/conman/exec exist
+      assert:
+        that:
+          - reg_dir_conman.stat.exists
+          - reg_dir_conman.stat.isdir
+      when:
+        - ansible_facts.os_family == "RedHat"
+        - ansible_facts.distribution_major_version == "8"
```
btravouillon, line 112: This test should also run for OpenSUSE.

btravouillon (approved):
Looks good to me. 👍

oxedions (approved):
Dear @osmocl 

Thanks a lot for this PR.
Ok for me, merging 😊

PR #469, opened 2020-11-12
Feat: addons/nic_nmcli - Merge routes task into nic_nmcli

oxedions:
Add routes handling in nmcli role, since routes are related to NIC here (not system routes).


oxedions:
Note that for the time being, I didn't found a simple way to remove routes (only remove all). So current role only add them. Removing has to be done manually.

oxedions:
Bug reported on VLANs. Investigating further.

PR #486, opened 2020-11-17
Feature/federate prom

Judunet:
New feature to include the multiple iceberg architecture.

### DIFF
--- a/roles/addons/prometheus_server/handlers/main.yml
+++ b/roles/addons/prometheus_server/handlers/main.yml
@@ -1,3 +1,13 @@
 ---
 - name: reload firewalld
   command: firewall-cmd --reload
+
+- name: service Restart prometheus server
+  service:
+    name: prometheus
+    state: restarted
+  #loop: "{{ dhcp_server_services_to_start }}"
+  #when:
+  #  - "'service' not in ansible_skip_tags"
+  #  - (start_services | bool)
+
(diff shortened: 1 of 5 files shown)


roles/addons/prometheus_server/tasks/main.yml
```diff
@@ -44,6 +44,20 @@
   tags:
     - firewall
 
+- name: "copy █ Install karma firewalld service configuration file"
```
btravouillon, line 47: I would copy both files alertmanager.xml and karma.xml in the same task, with a loop.

Off the top of my head:
```
copy:
  src: {{ item }}
  dest:  "/etc/firewalld/services/{{ item }}"
[...]
loop: ['alertmanager.xml', 'karma.xml']
```

roles/addons/prometheus_server/handlers/main.yml
```diff
@@ -1,3 +1,8 @@
 ---
 - name: reload firewalld
   command: firewall-cmd --reload
+
+- name: service Restart prometheus server
+  service:
+    name: prometheus
+    state: restarted
```
btravouillon, line 8: Can you add:
```
 when:
    - "'service' not in ansible_skip_tags"
    - (start_services | bool)
```

btravouillon (commented):
Hi @Judunet,

I started the review but have a question before going further. Is Prometheus federation compatible with Thanos?

btravouillon:
>  For Thanos i think it use federation, i will check that

Were you able to check that?

oxedions:
Dear @Judunet 

Any news on that matter ?

oxedions:
Closing for now.

Issue #488, opened 2020-11-18
log_client: allow more detailed rsyslog rules configuration


DilasserT:
Hello, 

It would be nice to have more flexibility with the rsyslog configuration. For example, I would like to change the forwarding rule: 

`
*.* @@{{networks[j2_node_main_network]['services_ip']['log_ip']}}:{{ log_client_server_port }}
`
to 
`
*.notice;kern.debug @@{{networks[j2_node_main_network]['services_ip']['log_ip']}}:{{ log_client_server_port }}
`

It can also be applied to the RULES section. 
There are a lot of possibilities, so it makes sense to me to allow the user to override the default template.
Also, if someone wants to tinker with one part while keeping the other part, it might be interesting to break down the configuration in several files ?

Something like that: 

/etc
├── rsyslog.conf
└── rsyslog.d
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;├── local-rules.conf
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;└── forwarding-rules.conf

What do you think ?

btravouillon:
I do agree.
Any user configuration must be kept in the user's inventory, so we need to find a way to override default templates with data from the inventory. This was done recently for [the new nhc role](https://github.com/bluebanquise/bluebanquise/pull/448#discussion_r520661613).

DilasserT:
About the default templates, would it be interesting to have that division ?
/etc
├── rsyslog.conf -> global rules
└── rsyslog.d
        ├── local-rules.conf -> node level rules
        └── forwarding-rules.conf -> forwarding rules

I am asking this because in my case, I would like to overwrite the forwarding rule while keeping the other configuration.

btravouillon:
I think so. How would you implement it in an ansible task?

DilasserT:
If I understand it correctly, in the nhc role there is an empty array for additional configuration files, a similar task can be implemented in log_client but with an array containing those files.

btravouillon:
Well, the issue with the implementation in the nhc role is that the custom files must be installed in the role directory, which is not in the scope of the user inventory: this will split the configuration reference between inventory and roles.

We could implement the code below:

```
- name: template █ Copy log custom files
  template:
    src: rsyslog.d_custom.conf.j2
    dest: "/etc/rsyslog.d/{{ item }}.conf"
    owner: root
    group: root
[...]

PR #494, opened 2020-11-25
Feat: core/log_client - Supersede default rsyslog configuration (#488)

DilasserT:


DilasserT:
Forgot to update changelog

roles/core/log_client/readme.rst
```diff
@@ -34,13 +34,62 @@ Log client verbosity defaults to info, it can be one of the following (defined i
...
+    #### MODULES ####
+
+    # The imjournal module bellow is now used as a message source instead of imuxsock.
+    $ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
+    $ModLoad imjournal # provides access to the systemd journal
+    #$ModLoad imklog # reads kernel messages (the same are read from journald)
+    #$ModLoad immark  # provides --MARK-- message capability
+    ...
+
+| You can also split the configuration into several files, using **log_client_custom_conf**.
```
btravouillon, line 61: What is the purpose of '|' at the beginning of each lines?

Also, if the variable for static configuration is **log_client_static_configuration**, I would name this one **log_client_custom_configuration**. That might also avoid conflict if one wants to write a custom.conf file.

roles/core/log_client/readme.rst
```diff
@@ -34,13 +34,62 @@ Log client verbosity defaults to info, it can be one of the following (defined i
...
+|
+| For example:
+
+.. code-block:: yaml
+
+  log_client_custom_conf:
+    - local_rules
+    - forwarding_rules
+
+  log_client_local_rules_rules: |
```
btravouillon, line 73: Might be **log_client_local_rules_conf**? (ok for the _conf here since the file will be local_rules.conf)

roles/core/log_client/templates/rsyslog.conf.j2
```diff
@@ -1,6 +1,13 @@
+#jinja2: lstrip_blocks: True
 #### Blue Banquise file ####
 ## {{ansible_managed}}
 
+{% if log_client_static_configuration is defined and log_client_configuration is not none %}
```
btravouillon, line 5: log_client_configuration ?

roles/core/log_client/templates/rsyslog.conf.j2
```diff
@@ -1,6 +1,13 @@
+#jinja2: lstrip_blocks: True
 #### Blue Banquise file ####
```
btravouillon, line 2: Can you use this PR to change Blue Banquise to BlueBanquise?

roles/core/log_client/tasks/main.yml
```diff
@@ -42,6 +42,17 @@
...
+- name: Add custom rsyslog files
+  template:
+    src: rsyslog.d_custom.conf.j2
+    dest: "/etc/rsyslog.d/{{ item }}.conf"
+    owner: root
+    group: root
+    mode: 0644
+  loop: "{{ log_client_custom_conf }}"
+  tags:
+    - custom_conf
```
btravouillon, line 54: @oxedions should the tag be 'template' here?

roles/core/log_client/readme.rst
DilasserT: Each "|" is used to start a newline, the default indentation did not feel right, should I remove them ?

roles/core/log_client/tasks/main.yml
oxedions: Yes please 🙂

DilasserT: It will be the same as the previous task, is it not a problem ?

oxedions: Both are templates, so it is ok.

roles/core/log_client/readme.rst
btravouillon: OK I got it. It looks weird when reading the readme in the terminal. If this is the same paragraph, I don't see no reason to force the newline, so please remove them.

oxedions:
Shouldn't we do the same than NHC role ?

Remove this:

```
{% if log_client_static_configuration is defined and log_client_static_configuration is not none %}

{{ log_client_static_configuration }}

{% else %}
```

And add a filter on the main configuration file template, to check if not defined in `log_client_configuration_files` ?

This is close to what is proposed here, but I think simpler.

DilasserT:
I am having trouble removing log_client_static_configuration:

In the nhc role, the configuration is written in /etc/nhc, but rsyslog expects /etc/rsyslog.conf.
Since the include directory must be defined in rsyslog.conf, maybe it can be added to the log_client_configuration_files ?
```
- name: template █ Install custom configuration files
  template:
    src: rsyslog.d_custom.conf.j2
    dest: "{{ item.path}}/{{ item.name }}"
    owner: root
    group: root
    mode: 0644
  loop: "{{ log_client_configuration_files }}"
  loop_control:
    label: "Install configuration file {{ item.name }}" 
```
```
log_client_configuration_files:
  - name: local-rules.conf
    path: /etc/rsyslog.d
    content: |
      my configuration for local-rules
      can be multiline
  - name: forwarding-rules.conf
    path: /etc/rsyslog.d
    content: |

      my configuration for forwarding-rules
  - name: rsyslog.conf
    path: /etc
    content: |
      # Include all config files in /etc/rsyslog.d/
      $IncludeConfig /etc/rsyslog.d/*.conf
```

btravouillon:
Good point.
I would default the path to */etc/rsyslog.d* to ease the configuration for end users:
```
template:
    src: rsyslog.d_custom.conf.j2
    dest: "{{ item.path | default('/etc/rsyslog.d') }}/{{ item.name }}"
```

```
log_client_configuration_files:
  - name: local-rules.conf
    content: |
      my configuration for local-rules
      can be multiline
  - name: forwarding-rules.conf
    content: |

      my configuration for forwarding-rules
```

Plus an explanation in the readme for those who want to skip the default template. Something like "You can override the default templating of /etc/rsyslog.conf by adding the configuration below to `log_client_configuration_files`:"

```
  - name: rsyslog.conf
    path: /etc
    content: |
      # Include all config files in /etc/rsyslog.d/
      $IncludeConfig /etc/rsyslog.d/*.conf
```

roles/core/log_client/readme.rst
```diff
@@ -34,13 +34,61 @@ Log client verbosity defaults to info, it can be one of the following (defined i
...
+- **content**: the content of your configuration file, it can be single line or multiline
+- **path**: the directory containing your configuration file, it defaults to /etc/rsyslog.d if it is not specified
+
+For example:
+
+.. code-block:: yaml
+
+  log_client_configuration_files:
+    - name: local-rules.conf
+      path: /etc/rsyslog.d
```
btravouillon, line 51: I would remove the path since this is the default. An example is available below with /etc.

roles/core/log_client/readme.rst
```diff
@@ -34,13 +34,61 @@ Log client verbosity defaults to info, it can be one of the following (defined i
...
+        ...
+
 To be done
 ^^^^^^^^^^
 
 Integrate journalctl logging instead of rsyslog.
 
 Changelog
 ^^^^^^^^^
+* 1.0.5: Enable custom rsyslog configuration with log_client_custom_config <dilassert@gmail.com>
```
btravouillon, line 91: This is a new feature, version should be 1.1.0. (https://semver.org/)

roles/core/log_client/tasks/main.yml
```diff
@@ -23,6 +23,7 @@
     - package
     - service
     - template
+    - custom_conf
```
btravouillon, line 26: This tag does not exist in the role.

roles/core/log_client/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 log_client_server_port: "{{ log_server_port | default(514) }}"
 log_client_verbosity: "info"
+log_client_configuration_files: []
```
btravouillon, line 4: This is unnecessary to declare this one as default, this is an optional parameter.

btravouillon (approved):
Looks good to me. 👍

oxedions (commented):
Dear @DilasserT 

Thanks a lot for this PR.
Ok for me 😊 

Merging

PR #496, opened 2020-11-26
Change /etc/conman.conf file mode to 600 to avoid security risks

gavillom:


oxedions:
Dear @gavillom 

Thanks a lot for this PR.

You are right, this need to be secured.
I will just test on my side that it works and then it is ok for me 😊

btravouillon:
The CI failed for OpenSUSE.

oxedions:
> 
> 
> The CI failed for OpenSUSE.

Yes, I have seen that. I wonder why. I will deploy an OpenSUSE to check.

btravouillon:
I suspect the conman service to run with a dedicated user (non root).

btravouillon:
That's it:
```
instance:/ # journalctl -u conman 
-- Logs begin at Thu 2020-11-26 18:56:45 UTC, end at Thu 2020-11-26 18:59:53 UTC. --
Nov 26 18:59:53 instance systemd[1]: Starting ConMan Console Management Daemon...
Nov 26 18:59:53 instance conmand[686]: ERROR:     Unable to open "/etc/conman.conf": Permission denied
Nov 26 18:59:53 instance systemd[1]: conman.service: Control process exited, code=exited status=1
Nov 26 18:59:53 instance systemd[1]: Failed to start ConMan Console Management Daemon.
Nov 26 18:59:53 instance systemd[1]: conman.service: Unit entered failed state.
Nov 26 18:59:53 instance systemd[1]: conman.service: Failed with result 'exit-code'.

instance:/ # cat /usr/lib/systemd/system/conman.service
[Unit]
Description=ConMan Console Management Daemon
After=network.target

[Service]
Type=forking
User=conman
Group=conman
ExecStart=/usr/sbin/conmand -c /etc/conman.conf

[Install]
WantedBy=multi-user.target

instance:/ # ls -l /etc/conman.conf
-rw------- 1 root root 328 Nov 26 18:59 /etc/conman.conf
```

oxedions:
It may be a good idea to run conman as user. I checked in RHEL 7, it was also running as root. But now that conman is no more provided by RedHat, nothing prevent us to change the way it works on RHEL 8, especially if it increase security.
Would worth a try.

@hmeAtos @bouriquet @osmocl would you have the appropriate hardware infrastructure to ensure it works, with SELinux enabled ? I do not have IPMI ready device right now (our Tumulus cluster is off due to IT room migration).

btravouillon:
> But now that conman is no more provided by RedHat, nothing prevent us to change the way it works on RHEL 8

How would you implement that? In the RPM or in the role?

oxedions:
I would to that in the rpm, since I also need to add the user inside the service file.

So an update in the rpm, and an update here in the task 🙂

btravouillon:
OK. If we should backport this security fix to 1.3, we would need to synchronize the publication of the RPM with the release of bluebanquise 1.3.2 to avoid permission denied for the existing user base (ansible sets ownership to conman while the service runs as root, or vice versa).

btravouillon:
Thinking about it, it might be easier to change this in the role only. This might work for both EL 7 and 8, whatever the vendor of the conman package.

btravouillon:
I had a doubt about /var/log/conman, so I checked with molecule:

```
[root@instance /]# systemctl status conman
● conman.service - ConMan Console Manager
   Loaded: loaded (/usr/lib/systemd/system/conman.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/conman.service.d
           └─user.conf
   Active: active (running) since Thu 2020-12-10 16:55:21 UTC; 3min 16s ago
     Docs: man:conmand(8)
 Main PID: 664 (conmand)
    Tasks: 1 (limit: 26213)
   Memory: 2.4M
   CGroup: /system.slice/docker.service/system.slice/conman.service
           └─664 /usr/sbin/conmand -P /var/run/conman/conmand.pid

Dec 10 16:55:21 instance conmand[664]: Open file limit set to 1048576
Dec 10 16:55:21 instance systemd[1]: Started ConMan Console Manager.
Dec 10 16:55:21 instance conmand[664]: Cannot create directory "/var/log/conman": Permission denied
Dec 10 16:55:21 instance conmand[664]: Unable to open logfile "/var/log/conman/instance.log": No such file or directory
```

The daemon can fail to create /var/log/conman but will not exit. I will fix this in a next update.

btravouillon:
BTW the verify test is not reliable...

```
TASK [Retrieve file /var/log/conman/instance.log system status] ****************
ok: [instance]

TASK [Assert log client file /var/log/conman/instance.log exists] **************
ok: [instance] => {
    "changed": false,
    "msg": "All assertions passed"
}
```

```
[root@instance /]# ls /var/log/conman/instance.log
ls: cannot access '/var/log/conman/instance.log': No such file or directory
```

roles/core/conman/tasks/main.yml
```diff
@@ -30,23 +30,66 @@
...
 - name: file █ Create /var/log/conman directory
   file: # noqa 208
     path: /var/log/conman
-    setype: conman_log_t
+    setype: "{{ (ansible_facts.selinux.status == 'enabled') | ternary('conman_log_t', omit) }}"
     state: directory
+    owner: conman
+    group: conman
     recurse: yes
+  notify: service █ Restart conman
```
oxedions, line 57: Is it relevant to restart conman here ?
This folder should be set at first deployment and is not expected to be removed later. Also, restarting now means conman will mostly try to start first time without the conman.conf generated because template is run later in the task.

roles/core/conman/tasks/main.yml
```diff
@@ -30,23 +30,66 @@
...
+    - ansible_facts.os_family == "RedHat"
+
+- name: copy █ Ensure service run with user conman
+  copy:
+    content: |
+      [Service]
+      User=conman
+      Group=conman
+      ExecStart=
+      ExecStart=/usr/sbin/conmand -P /var/run/conman/conmand.pid
```
oxedions, line 78: Why 2 ExecStart with first one empty ?

roles/core/conman/tasks/main.yml
```diff
@@ -30,23 +30,66 @@
...
+      PIDFile=/var/run/conman/conmand.pid
+    dest: /etc/systemd/system/conman.service.d/user.conf
```
oxedions, line 80: Why using a `conman.service.d/user.conf` instead of a simple `conman.service` ? Is there a need ? There will always be a single service for conman (or maybe I miss something).

oxedions (commented):
Dear @btravouillon 

Thanks a lot for code proposal.

Please see my comments 🙂

roles/core/conman/tasks/main.yml
btravouillon: This is explained in the commit message:  Override ExecStart with a new path for PIDFile (need to clear ExecStart first) owned by the conman user, this to allow the command to write the PIDFile.

ExecStart can be defined several time. To clear any previous value, we need to define it with an empty value.

roles/core/conman/tasks/main.yml
btravouillon: It's less intrusive to use the drop-in directory.

roles/core/conman/tasks/main.yml
btravouillon: Yes this is relevant. First deployment was months ago for some users of the stack.

btravouillon: > Also, restarting now means conman will mostly try to start first time without the conman.conf generated because template is run later in the task.

It does not restart "now". It notifies the handler. The restart will occur later, when the play will call the task `meta: flush_handlers`

oxedions:
Ok for me in this state for the 1.4.

When introducing new OS in 1.5, we will check how to proceed. 🙂

PR #497, opened 2020-11-26
feat: template custom NHC files from the inventory

btravouillon:
This change allows to define the name and content of each custom files
one want to install in /etc/nhc/ in the user inventory. This helps to
keep all the configuration in the inventory instead of splitting some
values in the inventory for templates and in the role for files.

### DIFF
--- a/roles/addons/nhc/tasks/main.yml
+++ b/roles/addons/nhc/tasks/main.yml
@@ -34,12 +34,16 @@
     owner: root
     group: root
     mode: 0644
+  when: "'nhc.conf' not in (nhc_files | default([]) | selectattr('name','defined') | map(attribute='name') | list)"
 
-- name: copy █ Copy nhc custom files
-  copy:
-    src: "{{ item }}"
-    dest: /etc/nhc/"{{ item }}"
+- name: template █ Copy nhc custom files
+  template:
+    src: nhc_custom_file.j2
+    dest: "/etc/nhc/{{ item.name }}"
     owner: root
     group: root
     mode: 0644
   loop: "{{ nhc_files }}"
+  loop_control:
+    label: "Copy file {{ item.name }}"
+  when: nhc_files is defined and nhc_files is iterable
--- a/roles/addons/nhc/defaults/main.yml
+++ b/roles/addons/nhc/defaults/main.yml
@@ -1,3 +1,2 @@
 ---
 nhc_custom_packages_to_install: []
-nhc_files: []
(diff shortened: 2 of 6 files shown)


roles/addons/nhc/tasks/main.yml
```diff
@@ -35,11 +35,14 @@
     group: root
     mode: 0644
```
oxedions, line 36: I think I would go even farther. Adding here a conditional to not generate nhc.conf if it is listed in the list above. So it allows advanced users to fully manipulate NHC.

```
when:
  - 'nhc.conf' not in (nhc_files | default([]) | selectattr('name','defined') | map(attribute='name') | list)
```

(Untested)

oxedions:
Dear @btravouillon 

Thanks a lot for this PR 🙂 
I added a single comment before reviewing full syntax.

roles/addons/nhc/tasks/main.yml
btravouillon: This task implements a similar mechanism to template nhc.conf with `nhc_static_configuration`, in a simpler way, documented in *General usage*.
I don't consider *nhc.conf* eligible to use in the task `template █ Copy nhc custom files` (documented in *Advanced usage). If one decide to do this, they will have an idempotency issue. To prevent this, I would tend to ensure `nhc.conf` is not allowed as a name in `nhc_files`.

oxedions: I didn't explained well what I want.

Since you implemented the capability to push whole files, using file name and content, I would like to:

1. Keep previous nhc.conf.j2 template, but remove the following lines inside it:

```
{% if nhc_static_configuration is defined and nhc_static_configuration is not none %}
{{ nhc_static_configuration }}
{% else %}
```

So template files allows simple configuration for simple usage, aka default NHC parameters and `nhc_checks` variable to fill it.

2. Add a conditional on this nhc.conf.j2 template, with an:

```
when:
  - 'nhc.conf' not in (nhc_files | default([]) | selectattr('name','defined') | map(attribute='name') | list)
```

To disable this template generation if `nhc.conf` is present in the `nhc_files` list. (so remove idempotent related issues).


This way, users have the choice, and role is simpler: they can either use the basic usage, with `nhc_checks`, or use a fully customized one with `nhc_files` (or combine both if desired).

btravouillon: Done

oxedions (approved):
Dear @btravouillon 

Thanks a lot for this PR.
This is perfect for me.

Merging.

PR #498, opened 2020-11-26
feat: add configuration of firewall zone to nic

btravouillon:
If you want to assign a firewalld zone to a network interface, define
the zone in the network linked to the interface. The zone name will be
added to the ZONE parameter of the ifcfg file.

As for other changes made by the nic role, the network service needs
to be restarted for changes to take effect.

### DIFF
--- a/roles/core/nic/readme.rst
+++ b/roles/core/nic/readme.rst
@@ -117,7 +117,18 @@ It is also possible to configure multiple ip per interface, using:
         - 172.16.0.3/16
         - 192.168.1.117/24
 
-MTU and/or Gateway can be set in the network file, and will be applyed to NIC linked to this network.
+To assign a network interface to a firewall zone:
+
+.. code-block:: yaml
+
+  network_interfaces:
+    - interface: ens5
+      ip4: 192.168.121.124
+      network: bb
+      zone: external
+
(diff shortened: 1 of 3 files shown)


btravouillon:
Result for a login node with backbone network interface ens5 connected to a network `bb` with a firewall zone `backbone`:

```
[root@login1 ~]# firewall-cmd --info-zone=backbone 
backbone (active)
  target: default
  icmp-block-inversion: no
  interfaces: ens5
  sources: 192.168.121.0/24
  services: ssh
  ports: 
  protocols: 
  masquerade: no
  forward-ports: 
  source-ports: 
  icmp-blocks: 
  rich rules: 
```

The `interface` is correctly added, but the `sources` still restrict access, which is the expected configuration from the role **firewall**.
Converting to Draft, I will update this PR to use a zone defined in `network_interfaces`.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/using-and-configuring-firewalld_configuring-and-managing-networking#using-zones-to-manage-incoming-traffic-depending-on-a-source_using-and-configuring-firewalld

oxedions (approved):
Dear @btravouillon 

This is ok for me, a must have. We will need to backport this into the futur nmcli role.

Merging.

PR #499, opened 2020-11-27
feat: ensure firewalld zones exist

btravouillon:
Allow to create a new zone to use with FirewallD.


oxedions (approved):
Dear @btravouillon 

Thanks a lot for this PR. It makes sense to have this in the firewall role.

Ok for me, merging !

PR #504, opened 2020-12-02
Feat: addons/nic_nmcli - Update routes, add zones

oxedions:
It is a feat, branch name mention fix, this is a mistake.

* Update routes to a much simpler format.
* Add Firewall zone.

Note that routes format change. Previous format was introduced recently, and should still not be used by any one, but in case of.

### DIFF
--- a/roles/addons/nic_nmcli/readme.rst
+++ b/roles/addons/nic_nmcli/readme.rst
@@ -10,6 +10,11 @@ The role also cover routes definitions on interfaces.
 This role provides all features availables in the main nmcli module.
 Please refer to `nmcli module documentation <https://docs.ansible.com/ansible/latest/collections/community/general/nmcli_module.html>`_ .
 
+.. warning:
+  This role needs **latest** nmcli.py module, from
+  https://github.com/ansible-collections/community.general/
+  ansible-collections:main branch.
+
 Instructions
 ^^^^^^^^^^^^
 
@@ -20,12 +25,14 @@ While all of the nmcli module options are supported,
 some provides more integrated features:
 
 * **conn_name**: is equal to **interface**, but has higher precedence over **interface** if both are set.
-* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname**  if both are set.
+* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname** if both are set.
 * **type**: is set to *ethernet* by default if not set.
-* **ip4**: can be set using a simple ipv4, then role will use **networks[item.network]['.prefix4']** or if not exist to **networks[item.network]['.prefix']** to complete address, or force address with prefix if string *'/'* is present.
+* **ip4**: can be set using a simple ipv4, then role will use **networks[item.network]['.prefix4']** or if not exist to **networks[item.network]['.prefix']** to complete address. You can force address with prefix if string *'/'* is present.
 * **mtu**: has higher precedence over **networks[item.network]['mtu']** if both are set.
-* **gw4**: has higher precedence over **networks[item.network]['.gateway4']** which has higher precedence over **networks[item.network]['.gateway']** (if set).
-* **routes4**: is a list, that defines routes to be set on the interface. See examples bellow.
+* **gw4**: has higher precedence over **networks[item.network]['.gateway4']** if set which has higher precedence over **networks[item.network]['.gateway']** if set. Note that gw4 is cannot be set at the same time than never_default4 (mutually exclusives).
+* **routes4**: is a list, that defines routes to be set on the interface. See examples bellow. Has higher precedence over **networks[item.network]['.routes4']** if set.
+* **route_metric4**: is to set general metric for gateway or routes (if not set on route level) for this interface. Has higher precedence over **networks[item.network]['.route_metric4']** if set.
+* **never_default4**: is related to ipv4.never-default nmcli parameter (DEFROUTE). Has higher precedence over **networks[item.network]['.never_default4']** if set.
 
 Basic ipv4
 """"""""""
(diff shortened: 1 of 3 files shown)


roles/addons/nic_nmcli/readme.rst
```diff
@@ -127,22 +134,17 @@ You can define routes at two levels:
...
-To remove a route later (here *10.12.0.0/24 10.10.0.2* on *enp0s8*), use the nmcli command this way:
-
-.. code-block:: text
+                - 10.11.0.0/24 10.10.0.2
+                - 10.12.0.0/24 10.10.0.2 300
 
-  nmcli connection show enp0s8 | grep ipv4.routes
-  nmcli connection modify enp0s8 -ipv4.routes "10.12.0.0/24 10.10.0.2"
+.. note:
+  In route4 list, first element is network to reach (0.0.0.0/0 for all), second
```
btravouillon, line 141: Ambiguous. Each element of the list is a tuple with the network destination in first position, gateway in second position and optionally the metric in third position.

roles/addons/nic_nmcli/readme.rst
```diff
@@ -20,12 +25,14 @@ While all of the nmcli module options are supported,
 some provides more integrated features:
 
 * **conn_name**: is equal to **interface**, but has higher precedence over **interface** if both are set.
-* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname**  if both are set.
+* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname** if both are set.
```
btravouillon, line 28: Could you wrap this part at 80 columns?

roles/addons/nic_nmcli/tasks/main.yml
```diff
@@ -65,7 +65,10 @@
     # {% endif %}
     # Allows to provide a simple ip4, or an ip4/prefix
     # Also support old prefix format (to be replaced by prefix4)
-    gw4: "{{ item.gw4 | default(networks[item.network].gateway4) | default(networks[item.network].gateway) | default(omit) }}"
+    gw4: "{{ (item.never_default4 | default(networks[item.network].never_default4) is not defined) | ternary( item.gw4 | default(networks[item.network].gateway4) | default(networks[item.network].gateway) | default(omit), omit) }}"
```
btravouillon, line 68: Could you add a comment that explain this test with never_default4? Or explain this in the commit message?
On first read, this is not obvious.

roles/addons/nic_nmcli/tasks/main.yml
```diff
@@ -65,7 +65,10 @@
...
+    routes4: "{{ (item.routes4 is defined or networks[item.network]['routes4'] is defined) |ternary( item.routes4 | default(networks[item.network]['routes4'] | default([]) ), omit) }}"
```
btravouillon, line 69: The `| default([])` seems unnecessary. When `(item.routes4 is defined or networks[item.network]['routes4'] is defined)` is true, one of both is defined, so `item.routes4 | default(networks[item.network]['routes4'])` should be sufficient. If `item.routes4` is not defined, we know that `networks[item.network]['routes4']` is. Or the first condition is false, then omit.
I wonder if you could achieve the same using a logic similar to route_metric4 below.
Am I missing something?

roles/addons/nic_nmcli/readme.rst
```diff
@@ -10,6 +10,11 @@ The role also cover routes definitions on interfaces.
 This role provides all features availables in the main nmcli module.
 Please refer to `nmcli module documentation <https://docs.ansible.com/ansible/latest/collections/community/general/nmcli_module.html>`_ .
 
+.. warning:
+  This role needs **latest** (2.0.0) nmcli.py module, from
+  https://github.com/ansible-collections/community.general/
+  ansible-collections:main branch.
+
```
btravouillon, line 17: The community.general 2.0.0 should be [available end of January 2021](https://mailchi.mp/redhat/the-bullhorn-15). In the meantime, I would not recommend to install a development version of this collection unless one really needs to use the new features introduced in this PR (routes and zones).
I would reflect this in the warning message: _If you plan to configure routes or zones with this role, you will have to install the development version of the community.general collection which is available at https://github.com/ansible-collections/community.general/ until the General Availability of the 2.0.0 release._

If bluebanquise 1.4 should land prior to this collection, I would indicate both features are tech-preview.

btravouillon (changes requested):

roles/addons/nic_nmcli/readme.rst
```diff
@@ -20,12 +25,14 @@ While all of the nmcli module options are supported,
 some provides more integrated features:
 
 * **conn_name**: is equal to **interface**, but has higher precedence over **interface** if both are set.
-* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname**  if both are set.
+* **ifname**: is equal to **physical_device**, but has higher precedence over **ifname** if both are set.
 * **type**: is set to *ethernet* by default if not set.
-* **ip4**: can be set using a simple ipv4, then role will use **networks[item.network]['.prefix4']** or if not exist to **networks[item.network]['.prefix']** to complete address, or force address with prefix if string *'/'* is present.
+* **ip4**: can be set using a simple ipv4, then role will use **networks[item.network]['.prefix4']** or if not exist to **networks[item.network]['.prefix']** to complete address. You can force address with prefix if string *'/'* is present.
```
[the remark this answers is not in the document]
oxedions: Updating.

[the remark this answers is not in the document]
oxedions: Good catch 😄

roles/addons/nic_nmcli/readme.rst
oxedions: Ok, I took your sentence.

roles/addons/nic_nmcli/tasks/main.yml
oxedions: Added comment in the task.

roles/addons/nic_nmcli/tasks/main.yml
oxedions: 😉 Good question: in fact, for an unknown reason, Ansible try to grab all values inside the ternary, the ones from true and the ones from false. So in ternary, you always need to provide default[] if you are not sure a value can exist.

roles/addons/nic_nmcli/readme.rst
oxedions: Updated the readme

oxedions:
Was used in production, and proved ok. Merging.

PR #506, opened 2020-12-10
fix: skip ansible-lint rule 208 when recurse: yes

btravouillon:
Remove mode: 0750 for file tasks where recurse: yes.

All the files in the directory will inherit this mode, which is wrong.
This mode was added in commit d70ad62 to resolve an error reported by
ansible-lint. Use #noqa 208 instead to skip the qa for this task, we can
rely on default mode set during the installation of the RPM.

### DIFF
--- a/roles/core/conman/tasks/main.yml
+++ b/roles/core/conman/tasks/main.yml
@@ -31,9 +31,8 @@
     - package
 
 - name: file █ Create /var/log/conman directory
-  file:
+  file: # noqa 208
     path: /var/log/conman
-    mode: 0750
     setype: conman_log_t
     state: directory
     recurse: yes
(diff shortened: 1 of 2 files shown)


btravouillon:
@giacomo-mcevoy if you want to review 😉

ginomcevoy:
@btravouillon thanks for the ping.
We tested this change in a virtual environment and now these two roles are idempotent.

One detail though, now a non-root user can do "ls /var/log/conman" and "ls /var/log/rsyslog" and see the directory contents (but cannot read the files). Not sure if this is acceptable or not.

btravouillon:
> One detail though, now a non-root user can do "ls /var/log/conman" and "ls /var/log/rsyslog" and see the directory contents (but cannot read the files). Not sure if this is acceptable or not.

I guess it is. These are the default permissions from the OS.

PR #511, opened 2020-12-20
ci: setup molecule for set_hostname

btravouillon:
The purpose of this test is to ensure the role can run, even if it's a single task role today.

The test does not change the hostname of the instance because it is not possible to change the hostname with hostnamectl in a running container (Could not set property: Failed to set static hostname: Device or resource busy).
Molecule, with the docker driver, will create a container with hostname `instance`, then will use this name as `inventory_hostname`. The role uses this variable in the task, which should not change the hostname. In the verify step, molecule gathers the facts and check that `ansible_hostname` is `instance`.


oxedions (approved):
Dear @btravouillon 

Thanks you for this PR.

This looks good to me. Merging.

Issue #482, opened 2020-11-16
[Multi-iceberg] Bug with pxe_stack role


Judunet:
Hi, 

I have a problem when i want to deploy multi-iceberg with the pxe_stack role.
I have configured like in documentation a nfs share for /etc/bluebanquise but when i want to deploy new node in iceberg2 (for example) it will delete the iceberg3 configuration.

May be a solution to resolve that is to move the pxe directory from /etc/bluebanquise.

Regards

oxedions:
Dear @Judunet 

Indeed, this is a bug in the design. I will think about that point, but clearly, the pxe folder should go somewhere else. It is probably time to create dedicated /etc folders for each tool (bootset and disklessset).

oxedions:
see related #484

oxedions:
Solved.

PR #514, opened 2021-01-07
Fix: core/pxe_stack - Update bootset configuration path

oxedions:
Fix issue described here: https://github.com/bluebanquise/bluebanquise/issues/482


johnnykeats (approved):
Deal

bouriquet (approved):

PR #517, opened 2021-01-20
Fix: core/pxe_stack - Fix bad hostname issue

oxedions:
RHEL 8.3 introduces change 
See https://bugzilla.redhat.com/show_bug.cgi?id=1881974


johnnykeats:
Why the sleep?

oxedions:
To ensure it spread over the whole system 🙂  Cost nothing

Issue #524, opened 2021-02-05
Customization of livenet image using playbook with connection chroot fails to match host


ginomcevoy:
When customizing a livenet image, the manual describes how to run a playbook using `connection: chroot` and the following hosts line:

```
hosts: /var/tmp/diskless/workdir/{{ image_name }}/mnt
```

This worked for the diskless manager in BlueBanquise release 1.3. However, for the new diskless manager, this fails to match the host:

```
[WARNING]: Could not match supplied host pattern, ignoring: /var/tmp/diskless/workdir/kitchen_image/mnt
```

This happens because a "/" has been added to the inventory file:

```
[root@management1 bluebanquise]# cat /var/tmp/diskless/workdir/kitchen_image/inventory/host
/var/tmp/diskless/workdir/kitchen_image/mnt/ ansible_connection=chroot
```

A PR will be proposed shortly to remove the extra '/' in the `host` inventory file.

oxedions:
Dear @ginomcevoy 

Thanks a lot for rising this issue, and for the associated PR 🙂

oxedions:
Closing since PR is merged.

Issue #523, opened 2021-02-05
sudo fails for diskless nodes with livenet image


ginomcevoy:
sudo fails on nodes booting a livenet image:

```
[vagrant@c002 ~]$ sudo su
sudo: /usr/bin/sudo must be owned by uid 0 and have the setuid bit set
```

Inspection of the squashfs image shows that the setuid 's' attribute is missing:

```
[root@management1 diskless]# chroot /var/tmp/diskless/workdir/my_image/mnt
[root@management1 /]# ls -lah /usr/bin/sudo
---x--x--x. 1 root root 162K Feb  3 13:54 /usr/bin/sudo
```

Fix:
During image creation, the contents are copied using `cp -r` instead of `cp -a`, the fix is to correct the cp flag. I will propose a PR shortly.

oxedions:
Dear @ginomcevoy 

Thanks a lot for rising this issue, and for the associated PR 🙂

oxedions:
Closing since PR is merged.

Issue #521, opened 2021-02-04
Cannot login or SSH after booting a livenet image with SELinux disabled


ginomcevoy:
After creating a livenet image with SELinux disabled (Activate SELinux = no in disklessset), the image boots but cannot login.
SSH fails and logging in via the console prompt fails without error (it prompts for login again).

Reviewing the console shows these kernel parameters:
```
[    0.000000] Command line: initrd=initramfs-kernel-4.18.0-193.el8.x86_64 root=live:http://10.11.0.1/preboot_execution_environment/diskless/images/kitchen_image/squashfs.img rw console=tty0 console=ttyS1,115200n8 nomodeset  rd.net.timeout.carrier=30 rd.net.timeout.ifup=60 rd.net.dhcp.retry=4 selinux=False 
```

Manually setting `selinux=0` instead of `selinux=False` in
`/var/www/html/preboot_execution_environment/diskless/images/<image>/boot.ipxe` corrects the issue and login is possible.

oxedions:
Dear @ginomcevoy 

Thanks a lot for rising this issue 🙂 

Would you accept to do a PR to fix the disklessset tool ?

Ox

oxedions:
Closing since PR is merged.

PR #525, opened 2021-02-05
Fix issues found when booting a diskless livenet image 

ginomcevoy:
This PR relates to the following issues regarding livenet images in the diskless role: #523 #521 #524

- For #523, using `cp -a` instead of `cp -r` will keep the setuid flag for /usr/bin/sudo.

- For #521, asking disklessset to disable SELinux would put `selinux=False` in the kernel command line, which would not disable SELinux during boot. The proposed fix will correctly write `selinux=0` in the kernel command line.

- For #524, the `self.MOUNT_DIRECTORY` variable is created with a `/` at the end and used in many places in the code. The proposed fix will keep the variable as-is but remove the `/` when writing the host file.

### DIFF
--- a/roles/addons/diskless/files/livenet_module.py
+++ b/roles/addons/diskless/files/livenet_module.py
@@ -221,8 +221,8 @@ class LivenetImage(Image):
         os.system('mount -o loop ' + self.IMAGE_DIRECTORY + 'tosquash/LiveOS/rootfs.img ' + self.MOUNT_DIRECTORY)
 
         # Put the operating system inside rootfs.img
-        logging.debug('Executing \'cp -r ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY + '\'')
-        os.system('cp -r ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY)
+        logging.debug('Executing \'cp -a ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY + '\'')
+        os.system('cp -a ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY)
 
         # Unmount rootfs.img file
         logging.debug('Executing \'umount ' + self.MOUNT_DIRECTORY + '\'')
--- a/roles/addons/diskless/files/livenet_module.py
+++ b/roles/addons/diskless/files/livenet_module.py
@@ -368,6 +368,7 @@ class LivenetImage(Image):
         os.mkdir(self.WORKING_DIRECTORY + 'inventory')
 
         # Create the ansible connection
+        # Use of [:-1] to remove last '/' from the MOUNT_DIRECTORY path
         logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
         os.system('echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
 
--- a/roles/addons/diskless/files/livenet_module.py
+++ b/roles/addons/diskless/files/livenet_module.py
@@ -368,8 +368,8 @@ class LivenetImage(Image):
         os.mkdir(self.WORKING_DIRECTORY + 'inventory')
 
         # Create the ansible connection
-        logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
-        os.system('echo \'' + self.MOUNT_DIRECTORY + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
+        logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
+        os.system('echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
 
         # Change image mountage status
         self.is_mounted = True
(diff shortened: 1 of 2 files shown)


oxedions (approved):
Dear @ginomcevoy 

Thanks a lot for this PR 😊 

It is ok for me !

I am requesting @pietersdavid approval also, since he is the one who made the new tool.

oxedions:
By the way, you are committing with @giacomo-mcevoy but doing PR with @ginomcevoy.
Is it on purpose ?

bouriquet (dismissed):

ginomcevoy:
> By the way, you are committing with @giacomo-mcevoy but doing PR with @ginomcevoy.
> Is it on purpose ?

No it is not, I will review this in my git config for next time.

btravouillon (changes requested):
Can you update the changelog?

ginomcevoy:
Hello, I have updated the changelog to mention this PR. Not sure I did the right thing with git, apologies if there is an inconvenience.

roles/addons/diskless/files/livenet_module.py
```diff
@@ -609,11 +609,14 @@ def generate_ipxe_boot_file(self):
...
 
+        # selinux should be 0 (False) or 1 (True)
+        image_selinux = '1' if self.selinux else '0'
+
         # Format image ipxe boot file template with image attributes
         file_content = self.__class__.get_boot_file_template().format(image_name=self.name,
                                                                       image_initramfs=self.image,
                                                                       image_kernel=self.kernel,
-                                                                      image_selinux=self.selinux)
+                                                                      image_selinux=image_selinux)
```
btravouillon, line 619: I would use `int(self.selinux)` and remove the lines 612-614 above.

roles/addons/diskless/files/livenet_module.py
```diff
@@ -221,8 +221,8 @@ def generate_squashfs(self):
         os.system('mount -o loop ' + self.IMAGE_DIRECTORY + 'tosquash/LiveOS/rootfs.img ' + self.MOUNT_DIRECTORY)
 
         # Put the operating system inside rootfs.img
-        logging.debug('Executing \'cp -r ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY + '\'')
-        os.system('cp -r ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY)
+        logging.debug('Executing \'cp -a ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY + '\'')
```
btravouillon, line 224: Why using `cp` instead of `rsync`?

roles/addons/diskless/files/livenet_module.py
```diff
@@ -221,8 +221,8 @@ def generate_squashfs(self):
...
+        os.system('cp -a ' + self.WORKING_DIRECTORY + 'generated_os/* ' + self.MOUNT_DIRECTORY)
```
btravouillon, line 225: Why use cp instead of rsync?
I would tend to use Python's `shutil.copytree()` instead of `os.system()`.

roles/addons/diskless/files/livenet_module.py
ginomcevoy: I have tested `int(self.selinux)` and it works, will add a commit soon.

roles/addons/diskless/files/livenet_module.py
ginomcevoy: Hello, this are good alternatives but I am not sure this is the right place for the change? NFS images also use `cp -a`.
Also I am not sure how to best use `shutil.copytree()` in this situation, it is known to cause problems when the destination directory already exists. For example I tried this:
```
shutil.copytree(self.WORKING_DIRECTORY + 'generated_os/', self.MOUNT_DIRECTORY)
```
And I get this error:
```
FileExistsError: [Errno 17] File exists: '/var/tmp/diskless/workdir/kitchen_image/mnt/'
```

roles/addons/diskless/files/livenet_module.py
btravouillon: It worked this way in 1.3. 😄

roles/addons/diskless/files/livenet_module.py
pietersdavid: The directory "self.WORKING_DIRECTORY + 'generated_os/" is not supposed to be already filled at this time, it is not a problem to use shutils.copytree() here.
I think use shutil.copytree() is recommanded here because it use directly Python commands instead of Linux system commands.

btravouillon: `dirs_exist_ok=True` should not raise an exception in your case.
Anyway, it is indeed not the right place to change this. I was just wondering why the use of `os.system()` instead of Python's built-in methods.

btravouillon (dismissed):
Looks good to me. 👍

bouriquet (dismissed):

pietersdavid (dismissed):

roles/addons/diskless/files/livenet_module.py
```diff
@@ -368,8 +368,8 @@ def mount(self):
         os.mkdir(self.WORKING_DIRECTORY + 'inventory')
 
         # Create the ansible connection
-        logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
-        os.system('echo \'' + self.MOUNT_DIRECTORY + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
+        logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
+        os.system('echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
```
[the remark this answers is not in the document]
oxedions: Yes, that would be nice 👍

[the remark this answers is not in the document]
oxedions:
```suggestion
        # Use [:-1] to remove last '/' from the MOUNT_DIRECTORY path
        logging.debug('Executing \'echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host\'')
        os.system('echo \'' + self.MOUNT_DIRECTORY[:-1] + ' ansible_connection=chroot\' > ' + self.WORKING_DIRECTORY + 'inventory/host')
```

oxedions:
The fail is not related to this PR. Merging. Thanks a lot for this PR 😉

PR #531, opened 2021-02-22
ci: update to ansible-lint v5

btravouillon:
The recent release of ansible-lint v5 brings some breaking changes.

Install yamllint and its configuration file at the root of the project.
Remove all the .yamllint conf files and run of the linter in molecule,
this is redundant with the static code analysis action.

Fix several issues reported by yamllint (indent, comment, trailing
spaces, new line char at EOF, etc.).

Explicitly define ANSIBLE_ROLES_PATH until ansible-lint#1375 is fixed.

Remove exclusion of rule E204 which was removed from ansible-lint rules.

PR #535, opened 2021-02-24
fix: always run the gather OS variable tasks

### LINKED ISSUE #532: roles pxe_stack and time is broken when using the skip-tags option
There is a bug in bluebanquise roles pxe_stack and time.

pxe_stack/tasks/main.yml:

name: include_vars ░ Gather OS specific variables
include_vars: "{{ item }}"
with_first_found:

files:
"vars/{{ ansible_facts.distribution | replace(' ','') }}{{ ansible_facts.distribution_major_version }}.yml"
"vars/{{ ansible_facts.os_family }}_{{ ansible_facts.distribution_major_version }}.yml"
"vars/{{ ansible_facts.distribution | replace(' ','_') }}.yml"
"vars/{{ ansible_facts.os_family }}.yml"
skip: true
tags:
firewall
internal
package
service
template

Some tags have been added to the task to allow (I suppose) these variables to be defined when executing pxe_stack role with some tags.

The problem is that when executing this role with a skip tag option to not execute package and service tasks. As a consequence, we do not include the needed variables and got an undefined variable.

btravouillon:
This avoid issues with --tags or --skip-tags.

Fix #532


btravouillon:
Failing test should be fixed with #531.

oxedions (approved):
Dear @btravouillon 

Thanks for this PR 🙂 

Ok for me, merging.

PR #539, opened 2021-03-11
ci: fix ansible versions and failing tests

btravouillon:
1) Update tox file:

   - The stack requires >= 2.9.13.
   - Don't test with ansible 2.8 anymore.
   - Run tests with ansible 2.9, 2.10 and 3.

2) Revert ebfa28f since ansible-base 2.10.7 is released (resolve `firewalld` to the correct FQCN).

3) Pin molecule CI for role time to Ubuntu 18.04 runner.

4) Ansible-lint 5.0.3 enforce rule unnamed-task for Task/Handler: meta.

Fix the failures below with ansible-lint:
```
unnamed-task: All tasks should be named
roles/addons/nic_nmcli/tasks/main.yml:120 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/addons/powerman/tasks/main.yml:22 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/addons/prometheus_server/tasks/main.yml:47 Task/Handler: meta  flush_handlers

role-name: Role name advanced-core does not match ``^[a-z][a-z0-9_]+$`` pattern
roles/advanced-core:0

unnamed-task: All tasks should be named
roles/advanced-core/advanced_dns_server/tasks/main.yml:103 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/conman/tasks/main.yml:105 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/dhcp_server/tasks/main.yml:85 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/dns_server/tasks/main.yml:118 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/firewall/tasks/RedHat.yml:22 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/log_client/tasks/main.yml:75 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/log_server/tasks/main.yml:121 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/nfs_server/tasks/main.yml:59 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
roles/core/pxe_stack/tasks/main.yml:268 Task/Handler: meta  flush_handlers

unnamed-task: All tasks should be named
[...]

PR #540, opened 2021-03-16
ci: update ansible deps in tox

btravouillon:
The stack requires >= 2.9.13.
Don't test with ansible 2.8 anymore.
Run tests with ansible 2.9, 2.10 and 3.

Also revert ebfa28f7 since ansible-base 2.10.7 released yesterday with the fix for the FQCN `firewalld`.


btravouillon:
Merge the change in #539

PR #542, opened 2021-03-30
Oxedions/dev/1.5/rhel like

oxedions:
Add compatibility for RHEL like distributions:

* Fedora 33
* RHEL 9 Alpha
* CentOS Stream 8
* OracleLinux 8
* CloudLinux 8

May worth improvement by merging some files in a future PR.


oxedions:
Time issue (whole matter in fact) must be resolved in another PR.

btravouillon:
> Time issue (whole matter in fact) must be resolved in another PR.

Can't you fix the time issue first then rebase this PR on top of it?

roles/core/pxe_stack/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 pxe_stack_enable_cloning: false
 pxe_stack_firewall_zone:
```
hmeAtos, line 3: not zone by default ?

roles/core/pxe_stack/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
...
+pxe_stack_sshd_force_PermitRootLogin: true
```
hmeAtos, line 4: why the `sshd_force_PermitRootLogin` is not an option on ssh_master configuration ?

oxedions:
> 
> 
> > Time issue (whole matter in fact) must be resolved in another PR.
> 
> Can't you fix the time issue first then rebase this PR on top of it?

I will. May take some time.

roles/core/pxe_stack/defaults/main.yml
oxedions: No, I added variable to make it more visible (I am using them a lot on Algoric, and found they were not enough visible and documented in the main documentation, I need to improve that).

roles/core/pxe_stack/files/osdeploy/redhat_9.ipxe
```diff
@@ -0,0 +1,38 @@
...
+
+imgstat
+
+echo | Booting in 4s ...
+echo |
+echo +----------------------------------------------------+
+
+sleep 4
+
+boot
```
hmeAtos, line 38: why not use `boot || shell`
it allows you to reset a dhcp + ipxe session if it didn't work the first time.
It's convenient for sys admins 😉

roles/core/pxe_stack/defaults/main.yml
oxedions: This parameter ensure you can ssh as root on the node after it has been deployed.
Recent versions of Fedora, and RHEL 9 do not allow ssh as root by default. So it is more during PXE 🙂

hmeAtos: ok, i didn't have this info, i'm ok then; )

roles/core/pxe_stack/defaults/main.yml
hmeAtos: ok

roles/core/access_control/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: include_tasks ░ Use OS dedicated task
-  include_tasks: "{{ (ansible_facts.os_family|lower|replace(' ','_')) }}_{{ ansible_facts.distribution_major_version }}.yml"
+  include_tasks: "{{ (ansible_facts.distribution|replace(' ','_')) }}_{{ ansible_facts.distribution_major_version }}.yml"
```
btravouillon, line 3: Why do you switch from `os_family` to `distribution`? It looks like all the distribution files are identical.

oxedions: It is a mistake. I will redo this whole PR, without editing the tasks code (except if needed).

Then I will propose another PR after to merge all things that could be from the same os_familly + major_version, to avoid having all these "same distributions" files.

roles/core/pxe_stack/files/osdeploy/redhat_9.ipxe
oxedions: Good point, I will try this and implement it in another PR

hmeAtos: ok thanks

PR #543, opened 2021-04-01
Feat: set_hostname - Add fqdn capability

oxedions:
Add FQDN capability for set_hostname.

Request made for an HPC cluster. Seems reasonable.

### DIFF
--- a/roles/core/set_hostname/defaults/main.yml
+++ b/roles/core/set_hostname/defaults/main.yml
@@ -1,2 +1,2 @@
 ---
-set_hostname_fully_qualified_domaine_name: false
+set_hostname_fqdn: false
--- a/roles/core/set_hostname/defaults/main.yml
+++ b/roles/core/set_hostname/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+set_hostname_fully_qualified_domaine_name: false
--- a/roles/core/set_hostname/molecule/default/converge.yml
+++ b/roles/core/set_hostname/molecule/default/converge.yml
@@ -1,6 +1,8 @@
 ---
 - name: Converge
   hosts: all
+  vars:
+    domain_name: bluebanquise.local
   tasks:
     - name: "Include set_hostname"
       include_role:
--- a/roles/core/set_hostname/readme.rst
+++ b/roles/core/set_hostname/readme.rst
@@ -14,6 +14,7 @@ NA.
 Changelog
 ^^^^^^^^^
 
+* 1.0.3: Add fully qualified domaine name capability. Benoit Leveugle <benoit.leveugle@gmail.com>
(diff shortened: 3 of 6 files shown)


hmeAtos:
ok for me, good feature  @oxedions

hmeAtos (commented):
have you write a documentation for set `set_hostname_fully_qualified_domaine_name` parameter ?

oxedions:
Good point :-)
On it

roles/core/set_hostname/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+set_hostname_fully_qualified_domaine_name: false
```
btravouillon, line 2: Could you define a shorter parameter name? `set_hostname_fqdn`?

roles/core/set_hostname/molecule/default/converge.yml
```diff
@@ -1,6 +1,8 @@
 ---
 - name: Converge
   hosts: all
+  vars:
+    domain_name: bluebanquise.local
```
btravouillon, line 5: Is `domain_name` a mandatory variable?

roles/core/set_hostname/readme.rst
```diff
@@ -14,6 +14,7 @@ NA.
 Changelog
 ^^^^^^^^^
 
+* 1.0.3: Add fully qualified domaine name capability. Benoit Leveugle <benoit.leveugle@gmail.com>
```
btravouillon, line 17: s/domaine/domain/

btravouillon (changes requested):

roles/core/set_hostname/defaults/main.yml
oxedions: Done 🙂

roles/core/set_hostname/molecule/default/converge.yml
oxedions: Yes, due to ternary 😐 Ansible always check all variables inside the ternary, even if not used.

roles/core/set_hostname/readme.rst
oxedions: Good catch, done.

roles/core/set_hostname/molecule/default/converge.yml
btravouillon: This needs to be documented then.

btravouillon (changes requested):
With this change the parameter `domain_name` must be defined. This needs to be documented somewhere.

btravouillon (approved):
Looks good to me. 👍

oxedions:
Thanks for reviewing.

PR #549, opened 2021-04-08
fix: add nodeset filter in powerman

btravouillon:
Add a new plugin filter to transform a list of nodes to a nodeset. Use
this filter in the powerman template.

### DIFF
--- a/roles/core/powerman/filter_plugins/nodeset.py
+++ b/roles/core/powerman/filter_plugins/nodeset.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleFilterError
+from ClusterShell.NodeSet import NodeSet
+
+def nodeset(nodes_list):
+    '''Convert a list of nodes to ClusterShell's NodeSet'''
+    nodeset = NodeSet(",".join(nodes_list))
+    return nodeset
+
+class FilterModule(object):
+    ''' NodeSet Jinja2 filter. '''
+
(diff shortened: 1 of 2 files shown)


btravouillon:
This is a proof of concept. If we take this path, this would need additional documentation, tests and some polish.
I would rather consider a common filter plugin instead of one dedicated to the powerman role.

oxedions:
Dear @btravouillon 

My apologies for the delay.
This is a very interesting dev. It would solve a lot of "unclean" templates.

And yes, I agree, I would also see a common filter.

But nodeset is so important, I even wonder if this could not include main Ansible. Did you test a PR on the Ansible github ?

btravouillon:
> But nodeset is so important, I even wonder if this could not include main Ansible. Did you test a PR on the Ansible github ?

It's a useful tool with our use cases in HPC and AI, but surely not for the majority of ansible users. Otherwise similar filter would have been implemented already.

oxedions:
Ok I see your pow.

I think having this filter in the stack would make sens, since many roles could benefit from it, and have way better final templates render (the slurm template is ugly if you have a lot of nodes).

My only concern is that filters means Python devs, and so need maintenance to keep up with Ansible 'API'. This filter is simple, and I think this is ok as long as filters remains simple.

oxedions:
Dear @btravouillon 

nodeset filter will be provided in next release into the whole stack: https://github.com/bluebanquise/tools/tree/main/bluebanquise_filters

So I can propose we merge this PR (allowing backport if needed), and once the nodeset rpm is only, we can remove the filter from the role and only keep the template.

Would that be ok for you ?

thiagocardozo (approved):

btravouillon:
Closed in favor of #596 .

PR #550, opened 2021-04-10
fix: use tag package for package tasks

btravouillon:
Was mistakenly replaced by always during move of the roles from addons
to core.


oxedions (approved):
Dear @btravouillon 

Good catch. Thanks a lot ! 😊 

Merging.

PR #551, opened 2021-04-13
conman: add conman user the authorization to write into /var/run/conman

MarbolanGos:
Hello,

I have found an issue with the role conman. It wants to start using the user conman but this user is not allowed to run in /var/run/conman because it does not have the access to this folder.

With the actual role when starting conman I had the error on the service:
`conmand[117848]: ERROR:     Unable to open pidfile "/var/run/conmand.pid": Permission denied`

In order to run the service correctly I had to do:
```
# rm -rf /var/run/conman
# mkdir /var/run/conman ; chown conman:conman /var/run/conman
# systemctl restart conman
```

The PR request attached adds the authorization for the user conman, group conman to use this /var/run/conman folder.


btravouillon:
Hello @MarbolanGos,

Can you provide additional insights about your environment?

The role ensures `/var/run/conman` is the home of user `conman` at: https://github.com/bluebanquise/bluebanquise/blob/48238747adb8398eb44916d938b0d4f4e6b6a61b/roles/core/conman/tasks/main.yml#L39-L47

I assume your OS is RHEL8. This task should create the `/var/run/conman` directory, which does not exist at first, with the correct permissions.

This role pass CI tests successfully, so I need more information to implement additional tests if we miss some standard use case here.

There is one case where it can fail, when I interactively create the directory as `root` user before run of the role. However, it's unlikely a user would create such directory outside of the role, without the correct permissions.

MarbolanGos:
Hi,
My OS is RHEL 8.3. The issue does not appear on running the role, it appears as you reboot. As this folder is tmpfs it is removed at reboot and then the service cannot use it anymore nor build it as was done by ansible role.

btravouillon:
Indeed... shame on me.
Let me review your change then. ;-)

roles/core/conman/tasks/main.yml
```diff
@@ -36,6 +36,16 @@
     state: present
     system: yes
 
+- name: user █ Ensure user conman can use /var/run/conman
+  copy:
+    content: |
+      D /var/run/conman 0775 conman conman -
+    dest: /usr/lib/tmpfiles.d/conman.conf
```
btravouillon, line 43: I would prefer `/etc/tmpfiles.d/conman.conf` (see [Configuration Directories and Precedence](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html#Configuration%20Directories%20and%20Precedence)), but are we the vendor or the local administrator here?
If there is a chance the upstream project (conman)  use this in a near future, usage of `/usr/lib/tmpfiles.d` may lead to some unexpected behavior.

btravouillon (commented):
Comments in line to open the discussion with reviewers.

roles/core/conman/tasks/main.yml
MarbolanGos: Sorry forgot this issue... I followed some RedHat blog page to design the role but I am not against your idea.

btravouillon (approved):
Looks good to me. 👍

oxedions (approved):
Dear @MarbolanGos 

Thanks a lot for this PR. 😊 

Merging

PR #556, opened 2021-04-28
nic: add ipv6_disable capability

MarbolanGos:
Hello,

In the case you need to disable IPv6, the nic role is not yet capable to do this. I have written a small patch to add this feature.

### DIFF
--- a/roles/core/nic/templates/ifcfg.j2
+++ b/roles/core/nic/templates/ifcfg.j2
@@ -89,3 +89,9 @@ MASTER={{ item.master }}
 {% if item.zone is defined and item.zone is not none %}
 ZONE={{ item.zone }}
 {% endif %}
+
+{# 9. Check if IPv6 has to be disabled #}
+{% if ipv6_disable is defined %}
+IPV6INIT=no
+IPV6_DISABLED=yes
+{% endif %}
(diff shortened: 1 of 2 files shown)


roles/core/nic/templates/ifcfg.j2
```diff
@@ -89,3 +89,9 @@ MASTER={{ item.master }}
 {% if item.zone is defined and item.zone is not none %}
 ZONE={{ item.zone }}
 {% endif %}
+
+{# 9. Check if IPv6 has to be disabled #}
+{% if ipv6_disable is defined %}
```
btravouillon, line 94: This test will match `ipv6_disable: false` and disable ipv6 for this interface. I would use `{% if ipv6_disable | default(false) | bool %}` instead.
Moreover, I believe this should be `item.ipv6_disable`.

btravouillon (changes requested):

roles/core/nic/templates/ifcfg.j2
oxedions: Dear @MarbolanGos

Any update on this PR ?

MarbolanGos:
As this role (nic) is deprecated by nic_nmcli and that this feature is available by default in the next default role. This PR is not relevant anymore.

oxedions:
Closing, as seen with @MarbolanGos

PR #559, opened 2021-04-30
Fix: kernel_config - Prevent crash if variable is None

oxedions:
Crash if variable is None. Small fix to prevent that.


btravouillon (approved):
Looks good to me. 👍

oxedions:
@btravouillon , Thanks a lot for the quick review 😊

PR #565, opened 2021-06-02
NFS_SERVER Fix bug with HA. the server var need to be an VIP address.…

Judunet:
Fix bug with nfs_server role:

Add new variable for conditional because when you want to add HA you need to set inn inventory/group_vars/all/general_settings/nfs.yml  the variable server with a virtual IP address and the role add a conditionnal with this. Create new variable "name_server" to specify on which server you want to add the export nfs and change the conditional in the task.


oxedions:
Directly seen with @Judunet 
In HA context, it is better to rely on PCS resource to export NFS

PR #566, opened 2021-06-04
Add conditionnal in nfs_client in src if value is an IP do not add ta…

Judunet:
Add conditionnal in nfs_client in src if value is an IP do not add take_over_network


oxedions (approved):
Dear @Judunet 

Thanks a lot for this PR. It makes sense. I tried to find a simpler code, but it seems this one is good.

Merging 😊

PR #568, opened 2021-07-02
Fix: log_server role remote logs improvement

thiagocardozo:
The default configuration has a few issues:
  - It replicates all logs from remotes in the server's
  - Separation by application instead of partition leads to too many files

This change proposes to store the remote logs with a similar configuration to non
remote logs.

*Emergency messages are still pushed for all to see.


thiagocardozo:
I`ve put the mngt node to also write local logs under rsyslog/mngt.

thiagocardozo:
Hello, everything should now be registered in rsyslog/$hostname.

oxedions:
Dear @thiagocardozo 

Yes, it works perfectly. Thanks a lot !

Ok for me, merging !

PR #577, opened 2021-07-07
Fix: disklessset - string used as a list

thiagocardozo:
Hotfix: Somehow in the backport a string was treated as a list. This is a fatal error.


oxedions (approved):
Dear @thiagocardozo 

Ok for me !

PR #578, opened 2021-07-09
Fix: nfs_server - exprt name consistency in readme

thiagocardozo:
The export /opt/software is used in mollecule tests and is also defined in resources/examples/simple_cluster/inventory/group_vars/all/general_settings/nfs.yml.

Change the nfs_server role readme to match that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Roles improvement or fix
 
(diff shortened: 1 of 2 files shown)


oxedions (commented):
Why closing?

thiagocardozo:
The actual request is to normalize the other way around, from /opt/software to /opt/softwares. If it's ok, I'll change the tests and the example to match the readme.

PR #584, opened 2021-07-21
Fix: conman - bad path for conman_execpath in RHEL8

oxedions:
Bad path for conman, simple fix.


gavillom:
Hi @oxedions.

Got the issue trying to install a cluster. Could be very interesting to have this fix available ;-)

oxedions:
Hello @gavillom 🙂 

Thanks for the feedback. Ok, I will merge it then.
I still do not understand why auto-test fails however, I will investigate.

PR #590, opened 2021-08-09
Fix: diskless - Correct python path

ginomcevoy:
The path for diskless files was incorrect, for example:

```
    - /lib/{{ diskless_python_version }}/site-packages/diskless
```

with diskless_python_version being "3.6" (RedHat 8) or "3.9" (RedHat 9). The path should be "python3.6" or "python3.9":

```
    - /lib/python{{ diskless_python_version }}/site-packages/diskless
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     - fix issues with dnf command in the livenet module (#528)
     - fix issues preventing access to nodes booting a livenet image (#525)
     - notify in readme/page that firewall does not work in chroot (#569)
+    - fix python path of diskless files (#590)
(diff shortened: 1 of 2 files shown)


oxedions:
Dear @ginomcevoy 

Yes, I know this bug. It will be solved by use of the bluebanquise/tools rpm, this is why I didn't fixed it in the dev branch.

See file https://github.com/bluebanquise/tools/blob/main/disklessset/disklessset.spec

But ok for this one, Lets merge it 😊

ginomcevoy:
OK yes you are right, the bluebanquise/tools rpm does the right thing. But It is still missing the diskless_parameters.yml file, I will soon create another PR in bluebanquise/tools for it.

PR #605, opened 2021-09-29
Fix: conman - force conman user gid and uid

oxedions:
Force conman user uid and gid, to prevent read/write issues in active passive HA context.


johnnykeats:
Since @neilmunday is working on Suse for next release, you should drop OpenSuse test for now.

PR #606, opened 2021-09-29
Update roles to work with OpenSuSE

neilmunday:
In `roles/core/repositories_server/tasks/main.yml` the `vars` file to use is determined by look at the Ansible facts for the OS. With this role the vars file for SUSE is called `SUSE.yml`. However, `ansible_facts.os_family` for OpenSuSE 15.3 returns `Suse` and thus the vars file is not matched due to the case.

This pull request fixes this issue by adding a search for the `ansible_facts.os_family` in upper case.


neilmunday:
Similar problem with the `dhcp_server` role. Last commit addresses this.

neilmunday:
Similar problem for `advanced_dhcp_server` role. Last commit addresses this.

neilmunday:
It should be noted that as it stands the two DHCP roles fail when trying to start the `dhcpd` service in OpenSuSE. This is because the role writes its main configuration file to `/etc/dhcp/dhcpd.conf` whereas the OpenSuSE set-up loads `/etc/dhcpd.conf`.

neilmunday:
Both DHCP roles fixed.

neilmunday:
`pxe_stack` role updated to work with OpenSuSE 15.3. Used @oxedions previous work on this to make it work. Tested successfully with a diskful install :-)

neilmunday:
Following discussions with @oxedions I will close this pull request and create a new one once the next big merge has been done.

PR #607, opened 2021-10-04
Feat: MLL - Add Minimal Live Linux

oxedions:


johnnykeats:
Why is it delayed?

oxedions:
If you have time, yes. I am having a hard time building MLL on each systems. Best thing would be to build the binaries on RockyLinux 8, and then dispatch these on each distributions containers to package them.

Issue #567, opened 2021-06-23
MI - SELinux boolean needed for HTTP over NFS export


oxedions:
Need 

`setsebool httpd_use_nfs on # this works`

To make http work on sub managements.

PR #620, opened 2021-10-22
Feat: nfs_client - Change sebooleans input variables

oxedions:
Fix issue https://github.com/bluebanquise/bluebanquise/issues/567 by adding a new variable able to handle multiple sebooleans.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@
 
 ### Breaking changes
 
+  - nfs_client: support multiple sebooleans (#620)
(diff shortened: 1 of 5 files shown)


oxedions:
Ready. @johnnykeats  please have a look.

johnnykeats:
Would be nice to provide backward compatibility, using `default` maybe?
Prevents bc.

oxedions:
Ok, I will have a look.

johnnykeats:
This does not really introduce a bc, as default already comply with previous default value.
This is a change, not a breaking change. You can merge it this way to me.

johnnykeats (approved):

oxedions:
Ok. I will review the whole example inventories anyway.

Issue #614, opened 2021-10-16
dns_server role reverse look-ups do not work


neilmunday:
Example:

IP subnet: 10.10.0.0/16

Generated named.conf:

```
zone "wolf.home.mundayweb.com" IN {
  type master;
  file "forward";
  allow-update { none; };
};

zone "in-addr.arpa" IN {
  type master;
  file "reverse";
  allow-update { none; };
};
```

Generared reverse file:

```
;#### Blue Banquise file ####
;## Ansible managed file, do not edit

$TTL 86400
$ORIGIN in-addr.arpa.
$INCLUDE "/var/named/reverse.soa"
@ IN NS wolf10.wolf.home.mundayweb.com.


2.0.10.10 IN PTR wolf1.wolf.home.mundayweb.com.
2.0.10.10 IN PTR wolf1-ice1-1.wolf.home.mundayweb.com.
3.0.10.10 IN PTR wolf2.wolf.home.mundayweb.com.
3.0.10.10 IN PTR wolf2-ice1-1.wolf.home.mundayweb.com.
10.0.10.10 IN PTR wolf10.wolf.home.mundayweb.com.
10.0.10.10 IN PTR wolf10-ice1-1.wolf.home.mundayweb.com.
1.0.10.10 IN PTR wolf0.wolf.home.mundayweb.com.
1.0.10.10 IN PTR wolf0-ice1-1.wolf.home.mundayweb.com.
```

`resolv.conf`:

```
# Generated by NetworkManager
search wolf.home.mundayweb.com
nameserver 10.10.0.10
```

nslookup test:

```
# nslookup wolf0.wolf.home.mundayweb.com
Server:         10.10.0.10
Address:        10.10.0.10#53

Name:   wolf0.wolf.home.mundayweb.com
Address: 10.10.0.1
```

Reverse test:

```
# nslookup 10.10.0.1
** server can't find 1.0.10.10.in-addr.arpa: NXDOMAIN
```

I can get reverse look-ups to work if I change the named.conf and reverse files to look like so:

named.conf:

```
zone "10.10.in-addr.arpa" IN {
  type master;
  file "/etc/named.d/reverse";
  allow-update { none; };
};
```

Reverse file:

```
;#### Blue Banquise file ####
;## Ansible managed file, do not edit

$TTL 86400
$ORIGIN 10.10.in-addr.arpa.
$INCLUDE "/etc/named.d/reverse.soa"
[...]

PR #621, opened 2021-10-23
Re-worked reverse zone file generation for dns_role to fix issue #614

neilmunday:
Changed the generation of the reverse zone file to instead be one file per network that is defined to be in DNS. The network prefix is used to generate the corresponding `in-addr.arpa` entries.

Reverse zones are named: `{{ network }}.rr.zone` and SOA files are similarly named `{{ network }}.rr.soa`

### DIFF
--- a/roles/core/dns_server/molecule/default/verify.yml
+++ b/roles/core/dns_server/molecule/default/verify.yml
@@ -10,7 +10,7 @@
       with_items:
         - /etc/named.conf
         - /var/named/forward.zone
-        - /var/named/reverse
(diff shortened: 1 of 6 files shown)


neilmunday:
Using a network with a prefix of 21 seems to create an invalid reverse zone config. Need to check and re-test.

neilmunday:
Fixed

johnnykeats:
@neilmunday 
Thanks, will have a look.

johnnykeats (approved):
Perfect, nice work!
Please add an entry in main CHANGELOG and ok to merge.

Issue #534, opened 2021-02-24
Distro repositories are not properly disabled on new version of Centos


gavillom:
Hi guys,

It feels like Centos 8 repositories are not properly disabled (with repositories_client_disable_distro_repos: true) with recent version of Centos.

It feels like repositories definition file name have been changed.

```
[root@gw yum.repos.d]# ll
total 48
-rw-r--r--. 1 root root  719 Nov 10 01:32 CentOS-Linux-AppStream.repo
-rw-r--r--. 1 root root  704 Nov 10 01:32 CentOS-Linux-BaseOS.repo
-rw-r--r--. 1 root root 1130 Nov 10 01:32 CentOS-Linux-ContinuousRelease.repo
-rw-r--r--. 1 root root  318 Nov 10 01:32 CentOS-Linux-Debuginfo.repo
-rw-r--r--. 1 root root  732 Nov 10 01:32 CentOS-Linux-Devel.repo
-rw-r--r--  1 root root  704 Feb 24 14:43 CentOS-Linux-Extras.repo
-rw-r--r--. 1 root root  719 Nov 10 01:32 CentOS-Linux-FastTrack.repo
-rw-r--r--. 1 root root  740 Nov 10 01:32 CentOS-Linux-HighAvailability.repo
-rw-r--r--. 1 root root  693 Nov 10 01:32 CentOS-Linux-Media.repo
-rw-r--r--. 1 root root  706 Nov 10 01:32 CentOS-Linux-Plus.repo
-rw-r--r--. 1 root root  724 Nov 10 01:32 CentOS-Linux-PowerTools.repo
-rw-r--r--. 1 root root  898 Nov 10 01:32 CentOS-Linux-Sources.repo
```

By default, _repositories_client_ role tries to remove a file named /etc/yum.repos.d/CentOS-Extras.repo (which is different than CentOS-Linux-Extras.repo).

Also BaseOS and AppStream file name has also been changed and so are still available when generating local repositories file.

I can propose a PR if you want.

Thanks

Mathis

btravouillon:
Indeed, files were renamed: https://git.centos.org/rpms/centos-repos/c/b759b17557b9577e8ea156740af0249ab1a22d70?branch=c8

If you open a PR, it would be nice to make the code backward compatible with CentOS < 8.3.

btravouillon:
[...]

PR #624, opened 2021-10-27
Fix for issue 534

neilmunday:
Potential fix for issue #534.

Maintains backwards compatibility for CentOS 8.3 and below as well as forwards compatibility for CentOS 8.4 onwards.


oxedions:
Dear @neilmunday 

I think we can assume that users will now be up to date. So you can create a single file for CentOS_8.
Also, since CentOS is near to deprecation and end of story, I think this is more than enough. 😊

neilmunday:
@oxedions done. Hopefully this will pass the tests now.

johnnykeats:
Will test this code into a dedicated container to validate it.

johnnykeats:
Tested and confirm it works with this fix. This PR is ok for me, and the molecule test broken.
Mostly container image not up to date.

neilmunday:
@johnnykeats CHANGELOG updated though conflicts with the master branch version will need fixing.

oxedions:
Merging. Thanks a lot @neilmunday  for this fix 😊

PR #627, opened 2021-11-02
Add OpenSUSE 15 support

neilmunday:
This pull request adds OpenSUSE 15 support.

Note: the following OpenSUSE repositories are required:

- backports-update
- bluebanquise
- non-oss
- os
- oss
- sle-update
- update
- update-non-oss

The _os_ repository is the OpenSUSE DVD contents.

The _bluebanquise_ repository contains the following packages:

- bluebanquise_filters-1.0.1-1.noarch.rpm
- ipxe-x86_64-bluebanquise-1.3.0-1.lp153.noarch.rpm

I have tested the roles on a OpenSUSE management node and OpenSUSE compute node. Additionally I have tested that an OpenSUSE management node can deploy a CentOS host and vice-versa. Note: all VMs, not real hosts.


johnnykeats:
An impressive work! 🧙‍♂️

Will check all in VMs.

First note:
Some feature are missing in the autoyast.xml file ( -> @neilmunday ).
These features were introduced for RHEL a long time ago, and were even not backported into Ubuntu code. ( -> @oxedions )

* ep_preserve_efi_first_boot_device
* ep_autoinstall_pre_script
* ep_autoinstall_post_script

While checking code, found that EFI script in DGX curtin file is better than the one used in kickstart file, could be an opportunity to update that point.

oxedions:
You are right, I completely forgot these ones 😆 

I found the DGX efi line while fighting with Curtin :
`efibootmgr -o $(efibootmgr | grep BootCurrent | cut -d" " -f2),$(efibootmgr | grep BootOrder | sed "s/BootOrder:\ //" | sed "s/$(efibootmgr | grep BootCurrent | cut -d" " -f2),//")`

This line replaces the whole post/pre script things in the current kickstart. I will update it and also add these features into Ubuntu code.

neilmunday:
@johnnykeats thanks for the pointers. I've updated the `pxe_stack` role to incorporate these. In doing so I was also able to find a `validate` command for AutoYast XML files which was handy and this is now included also.

johnnykeats:
@neilmunday Thanks

Will test it on a Suse Enterprise system and let you know asap.

neilmunday:
Working on adding SLES 12 support also...

neilmunday:
Some SLES 12 SP5 notes:

- Tried using Ansible 2.8.1 (from SUSE Package Hub) with Python 2.7 but nmcli module has a bug: https://bugzilla.suse.com/show_bug.cgi?id=1194718 and the this version of Ansible doesn't work with the SSH roles.
- Switched to using Python 3.6 and Ansible 2.11.7 from PyPi however, hit this bug with the Ansible nmcli plugin: https://github.com/ansible-collections/community.general/issues/3948. Can work round this by editing /root/.local/lib/python3.6/site-packages/ansible_collections/community/general/plugins/modules/net_tools/nmcli.py and /root/.local/lib/python3.6/site-packages/ansible_collections/community/general/plugins/modules/nmcli.py and comment all references to: `routing-rules`

neilmunday:
Tested set-up of SLES 12 SP5 management node and the deployment of a SLES 12 SP5 node. This works but the NetworkManager RPM is not on the SLES 12 ISO so the OS gets installed with Wicked. On OpenSUSE 15 (and I therefore assume SLES 15 also) it is possible to set the use of NetworkManager via autoyast. When running the `nic_nmcli` role on the deployed host the connection will drop due to `wicked` being stopped. Even with NetworkManager then installed manually etc. the role fails with:

```
Error: mandatory 'ifname' not seen before 'connection.interface-name'.
```

This appears to be due to the interface being named `Wired Connection 1` on the deployed host. If I rename the this connection to `eth0` then the role then works, e.g.

```
wolf1:~ # nmcli c
NAME                UUID                                  TYPE            DEVICE
Wired connection 1  64e2de50-66b6-4204-a323-151fbc66bdd1  802-3-ethernet  eth0
wolf1:~ # nmcli c mod 64e2de50-66b6-4204-a323-151fbc66bdd1 connection.id eth0
wolf1:~ # nmcli c
NAME  UUID                                  TYPE            DEVICE
eth0  64e2de50-66b6-4204-a323-151fbc66bdd1  802-3-ethernet  eth0
```

Not sure how to work around this at the moment. If I can easily add the SDK repository to the autoyast deployment then in theory NetworkManager could be set to be used at the deployment phase.

neilmunday:
Success! Adding the SDK repo to the Autoyast file and enabling NetworkManager here etc. results in a SLES 12 host deployed with NetworkManager working :-).

Next step is to test SLES 15 SP3...

neilmunday:
Ansible 2.9 in SLES 15 SP3 suffers from a known `nmcli` bug. I've raised this on the SUSE bug tracker: https://bugzilla.suse.com/show_bug.cgi?id=1195078

neilmunday:
Tested deploying a SLES 15 node (VM) ok :-)

oxedions:
I tested all roles, it works fine. I will update advanced_roles later.

Thanks a lot @neilmunday  for this impressive PR 😊 

So next step is to allow bootset to handle htdocs path as needed.

Merging the PR !!

oxedions:
Many thanks to you !! This is a superb PR !!! 😊

PR #632, opened 2021-11-09
Fix: Tox - Update envlist

oxedions:
I think we should restrain python 2.7 to Ansible 2.9 as Rhel7 is the only distrib we are supporting with python 2.7, and it is based on Ansible 2.9.
I see always too much issues with more recent Ansible, event if it should be supported by Ansible team.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,3}-ansible{29,210,3}
+envlist = py27-ansible29,py3-ansible{29,210,3}
 skipsdist = true
 
 [testenv]


johnnykeats:
What about Ubuntu 18.04?

oxedions:
I will check. To be honest, I am not sure we should support 18.04. We have 20.04 and it works well.
18.04 is a pain on some recent hardware.

johnnykeats:
```
root@019b092f9e9e:/# ansible --version
ansible 2.5.1
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.17 (default, Feb 27 2021, 15:10:58) [GCC 7.5.0]
root@019b092f9e9e:/# 
```

Even lower on 18.04.

btravouillon:
Note that Ansible 2.9 EOS will be May 23, 2022: https://groups.google.com/g/ansible-announce/c/kegIH5_okmg/?pli=1
FWIW I run a cluster of Ubuntu 18.04 hosts. Nonetheless, our Ansible controller is a container w/ Python3 + Ansible 2.9.

You should definitely drop support of Python 2 on the Ansible controller.

oxedions:
Ok, so this converge to 2 things:

- We should really consider dropping Python 2 (as stated by @btravouillon, a container is the perfect solution)
- We need to keep Ubuntu 18.04 support for now, as I also had another user asking to keep 18.04 in the mean time.

And I didn't realized that Ansible 2.9 EOL was so close. We need to consider migrating to upper versions with 1.6 version of BB.

johnnykeats (approved):

btravouillon (approved):
Looks good to me. 👍

oxedions:
Thanks for having a look. Merging.

Issue #557, opened 2021-04-28
ssh_master: host key mismatch log level


MarbolanGos:
Hello,

In the case of diskless nodes, after every reboot, the `known_hosts` file is incorrect thus when logging in the node, there are some warning displays. This can be disturbing to get for each node such a warning.

In order to fix this, I have added this:
```
--- roles/core/ssh_master/templates/config.j2   2021-04-28 08:54:18.527323434 +0300
+++ roles/custom/ssh_master/templates/config.j2 2021-04-20 10:42:20.733492860 +0300
@@ -11,6 +11,7 @@ Host {{ host }}
           security.ssh.hostkey_checking == false %}
     StrictHostKeyChecking no
     UserKnownHostsFile=/dev/null
+    LogLevel QUIET
     {% endif %}
     {% if ssh_jump is none %}
     Hostname %h-{{ macro_node_main_network }}
```

What I wanted at the beggining was to be able to have security checkups still being active when being on a diskfull node and removing host checking when not being but I was not able to do it easily. 

This issue is a cosmetic issue but can help administrating the servers.

oxedions:
Fixed in PR #648 😊

Issue #537, opened 2021-03-08
hostkey_checking is global 


yhenriet:
Hello, 
It seems that the hostkey settings is set globaly for every host
from inventory/group_vars/all/general_settings/security.yml:
```
---
security:
  ssh:
    hostkey_checking: true # enable or disable hostkey checking
```

We can add the var to the host , but the role ssh_master is not using it in this way :

From roles/core/ssh_master/templates/config.j2 : 

```
Host {{ host }}
    {% if **security.ssh.hostkey**_checking is not defined or
          security.ssh.hostkey_checking == false %}
    StrictHostKeyChecking no
    UserKnownHostsFile=/dev/null
    {% endif %}
```
So we can't only define the host check for only one host for example . 

But may be we are thinking wrong :)

btravouillon:
Hi @yhenriet ,

Would the change at https://github.com/bluebanquise/bluebanquise/compare/master...btravouillon:fix/537 address your need?

oxedions:
It should work:

```
{% macro write_host(host,macro_node_main_network,ssh_jump) %}
Host {{ host }}
    {% if hostvars[host].ssh_master_stricthostkeychecking is defined %}
    StrictHostKeyChecking {{ hostvars[host].ssh_master_stricthostkeychecking }}
    {% endif %}
    {% if hostvars[host].ssh_master_userknownhostsfile is defined %}
    UserKnownHostsFile {{ hostvars[host].ssh_master_userknownhostsfile }}
```

Will apply to hosts that you desire to exclude from the global configuration, while the last part:

```
Host *
{% if security.ssh.hostkey_checking is not defined or
      security.ssh.hostkey_checking == false %}
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
{% endif %}
```

only apply to all hosts that do not have these parameters set. I just do not know if ssh when reading this file will precedence `Host myhost` parameters over `Host *`. I assume @btravouillon already tested and validated it.

btravouillon:
> I"m trying to avoid the message in clush for the computes nodes . So to not have the checking with compute , but to keep it for others .

What message are you trying to avoid?
Note that for production system, `StrictHostKeyChecking` should never be set to `no`.

`security.ssh.hostkey_checking` is indeed a parameter of the target host where the role `ssh_master` run, to enable or disable host key checking for all hosts configured in `~/.ssh/config` of the target host. We won't change the behavior, this would introduce a breaking change.

[...]

Issue #418, opened 2020-10-15
ssh_master : template █ Generate /root/.ssh/config fail


btravouillon:
# Description

The playbook fails at task `ssh_master : template █ Generate /root/.ssh/config`:

```
fatal: [management1]: FAILED! => changed=false
  msg: |-
    AnsibleError: template error while templating string: expected token '=', got '.'. String:
[...]
```

This is a fresh deployment with bluebanquise-vagrant. I'm running ansible for the first time to configure the management node.

# Info

OS: CentOS 7.8
Branch: master

# Reproducer

My playbook:
```
# cat playbooks/managements.yml
---
- name: managements playbook
  hosts: "management1"
  roles:
    - role: bluebanquise
      tags: bluebanquise

- name: managements playbook
  hosts: "mg_managements"
  roles:

    - role: set_hostname
      tags: set_hostname
    - role: nic
      tags: nic
    - role: repositories_server
      tags: repositories_server
    - role: repositories_client
      tags: repositories_client
    - role: firewall
      tags: firewall
    - role: hosts_file
      tags: hosts_file
    - role: ssh_master
      tags: ssh_master
    - role: dhcp_server
      tags: dhcp_server
    - role: advanced_dns_server
      tags: ['dns', 'dns_server']
    - role: dns_client
      tags: ['dns', 'dns_client']
    - role: time
      tags: time
      vars:
        time_profile: server
    - role: pxe_stack
      tags: pxe_stack
    - role: nfs_server
      tags: nfs_server

  tasks:
    - package:
        name: vim
        state: present
```

Command:
```
[root@management1 bluebanquise]# ansible-playbook playbooks/managements.yml -l management1
[...]
TASK [ssh_master : template █ Generate /root/.ssh/config] *****************************************************
Thursday 15 October 2020  22:10:37 +0000 (0:00:00.516)       0:00:09.047 ******
fatal: [management1]: FAILED! => changed=false
  msg: |-
    AnsibleError: template error while templating string: expected token '=', got '.'. String:
    {% macro node_main_network(macro_network_interfaces,macro_current_iceberg_network) %}
    {{ macro_network_interfaces | selectattr('network','defined') | selectattr('network','match','^'+macro_current_iceberg_network+'-[a-zA-Z0-9]+') | map(attribute='network') | list | first }}
    {% endmacro %}

    {# Macro to avoid redundancy #}
    {% macro write_host(host,macro_node_main_network,ssh_jump) %}
[...]

PR #638, opened 2021-11-18
Oxedions/ssh logs

oxedions:
Attempt to solve issues #557 and #537 
Will also consider #418


oxedions:
Dear @ginomcevoy 

You are completely right, my bad. I fixed the readme.

PR #648, opened 2021-12-14
Fix: bluebanquise - Add missing bluebanquise_filter package

oxedions:
Add missing filters


oxedions:
I know this one is broken, will be fixed later.

PR #649, opened 2021-12-14
Fix: pxe_stack - Add bootset as package

oxedions:
Replace template by package.

Should have been made a long time ago


oxedions:
Auto test will be fixed later

PR #662, opened 2022-01-11
Fix: nic_nmcli - add missing register of nic_nmcli_apply variable

ginomcevoy:
The optional parameter of nic_nmcli role does not work, the role fails with this error message:

```
# ansible-playbook computes.yml --tags nic_nmcli -e nic_nmcli_force_nic_restart=true
...
TASK [nic_nmcli : shell █ Force apply nic settings] **********************************************************************
Tuesday 11 January 2022 16:31:38 +0100 (0:00:01.108) 0:00:08.516 *******
fatal: [management1]: FAILED! =>
msg: '''nic_nmcli_apply'' is undefined
```

This PR adds the missing variable definition, and the role works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     - add ip4_manual entry (#618)
     - add dns4 and dns4_search vars logic (#585)
     - improve role capabilities (#558)
+    - add missing register of nic_nmcli_apply variable (#662)
   - powerman:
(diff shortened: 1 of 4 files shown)


oxedions (approved):
Dear @ginomcevoy 

Thanks a lot for reporting and fixing this bug 😊 

Ok for me, merging.

PR #587, opened 2021-08-05
Fix: pxe_stack - Force substitution from file to symlink

thiagocardozo:
In the upgrade from BB v1.4 to v1.5, some files became links. Without the "force" attribute, Ansible refuses to do the replacement.

```
TASK [pxe_stack : file █ Generate links for all supported OS] ****************************************************************************************************************************************
Thursday 05 August 2021  16:10:34 -0300 (0:00:27.997)       0:01:47.544 *******
skipping: [mngt0-1] => (item=[{'major': 7, 'distributions': ['redhat', 'centos']}, 'redhat'])
failed: [mngt0-1] (item=[{'major': 7, 'distributions': ['redhat', 'centos']}, 'centos']) => changed=false
  ansible_loop_var: item
  gid: 48
  group: apache
  item:
  - distributions:
    - redhat
    - centos
    major: 7
  - centos
  mode: '0644'
  msg: refusing to convert from file to symlink for /var/www/html/preboot_execution_environment/osdeploy/centos_7.ipxe
  owner: apache
  path: /var/www/html/preboot_execution_environment/osdeploy/centos_7.ipxe
  secontext: unconfined_u:object_r:httpd_sys_content_t:s0
  size: 1367
  state: file
  uid: 48
skipping: [mngt0-1] => (item=[{'major': 8, 'distributions': ['redhat', 'centos', 'cloudlinux', 'oraclelinux', 'almalinux', 'rocky']}, 'redhat'])
failed: [mngt0-1] (item=[{'major': 8, 'distributions': ['redhat', 'centos', 'cloudlinux', 'oraclelinux', 'almalinux', 'rocky']}, 'centos']) => changed=false
  ansible_loop_var: item
  gid: 48
  group: apache
  item:
  - distributions:
    - redhat
    - centos
    - cloudlinux
    - oraclelinux
    - almalinux
    - rocky
    major: 8
  - centos
  mode: '0644'
  msg: refusing to convert from file to symlink for /var/www/html/preboot_execution_environment/osdeploy/centos_8.ipxe
  owner: apache
  path: /var/www/html/preboot_execution_environment/osdeploy/centos_8.ipxe
  secontext: unconfined_u:object_r:httpd_sys_content_t:s0
  size: 1374
  state: file
  uid: 48
```

PR #666, opened 2022-01-20
Fix: pxe_stack force substitution from file to symlink

thiagocardozo:
Had the same problem as described in [#587](https://github.com/bluebanquise/bluebanquise/pull/587).
This time from bb-1.4 to 1.5.

Somehow the fix was removed during the update.


oxedions:
Dear @thiagocardozo 
Thanks a lot ! You are right, I do not know but it was erased. Thanks for the fix 😊

Issue #682, opened 2022-03-02
[BUG] dns_server: reverse resolution not working for /22 network prefix


ginomcevoy:
This is relevant to the dns_server core role, currently it does not configure certain networks correctly and reverse resolution fails.

## Context

When defining a network with the /22 prefix (255.255.252.0 netmask), there can be nodes that are outside the "C-class range" of the network. For example, the 192.168.100.0/22 network can go up to 192.168.103.255, so it is possible to have a node with the 192.168.103.1 IP address inside that network.

The issue is that this node is not correctly configured in the dns_server role, namely the reverse resolution fails.

## Process

1. Define this network 192.168.100.0/22 in network.yml:
```
  ice1-2:
    subnet: 192.168.100.0
    prefix: 22
    netmask: 255.255.252.0
    broadcast: 192.168.103.255
    is_in_dns: true
```
2. Assign the interface of a node to 192.168.103.1 which is inside this ice1-2 network (login1 in this example). 

3. Run dns_server role, e.g.
```
ansible-playbook playbooks/managements.yml -t dns_server
```

## Expected result

Both forward and reverse DNS resolutions should work:
```
[root@management1 bluebanquise]# host login1
login1.tumulus.local has address 192.168.103.251

[root@management1 bluebanquise]# host 192.168.103.251
251.103.168.192.in-addr.arpa domain name pointer login1-ice1-2.tumulus.local.
251.103.168.192.in-addr.arpa domain name pointer login1.tumulus.local.
```
## Current result

DNS resolution works, but reverse resolution fails:
```
[root@management1 bluebanquise]# host login1
login1.smc.local has address 192.168.103.251
[root@management1 bluebanquise]# host 192.168.103.251
Host 251.103.168.192.in-addr.arpa not found: 5(REFUSED)
```
In fact, the reverse address has not been calculated properly (notice 192.168.100 instead of 192.168.103):
```
[root@management1 bluebanquise]# host 192.168.100.251
251.100.168.192.in-addr.arpa domain name pointer login1-ice1-2.tumulus.local.
251.100.168.192.in-addr.arpa domain name pointer login1.tumulus.local.
```
## Possible Fix

Not a DNS guru by any means, but I saw articles regarding this issue, and the recommendation (or standard practice?) is to "delegate multiple zones of the closest smaller size", e.g.

[...]

PR #683, opened 2022-03-03
Fix for issue #682 in dns_server role

neilmunday:
This mod changes the `dns_server` role so that reverse zone files are created in the form `xxx.yyy.zzz.rr.zone` for every host/BMC IP that matches.

If this is acceptable, the same mods can be made to to the `advanced_dns_server` role.

### DIFF
--- a/resources/documentation/conf.py
+++ b/resources/documentation/conf.py
@@ -86,7 +86,7 @@ pygments_style = None
 html_logo = "logo.png"
 html_theme = "sphinx_rtd_theme"
 html_theme_path = ["_themes", ]
-html_css_files = ["custom.css",]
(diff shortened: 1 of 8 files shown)


oxedions:
Dear @neilmunday 

Thanks a lot for this fix. I need some time to investigate this, but will come back to you soon 😊

ginomcevoy:
As commented in the issue, I did some manual checks and it looks OK to me. But yes please also fix advanced_dhcp_server role when the solution is approved.

oxedions:
My apologies for the delay. I was focusing on packages process to include open SUSE.

This is also ok for me. Merging, thanks a lot @neilmunday 😊

PR #687, opened 2022-04-01
Feat: pxe_stack - Allow root or sudouser

oxedions:
Allow to choose between root or sudo user when deploying OS.

Default now goes to sudo user.

### DIFF
--- a/roles/core/pxe_stack/tasks/client_os/Suse.yml
+++ b/roles/core/pxe_stack/tasks/client_os/Suse.yml
@@ -19,7 +19,7 @@
     mode: 0644
     owner: "{{ pxe_stack_apache_user }}"
     group: "{{ pxe_stack_apache_group }}"
-    #validate: /usr/bin/xmllint --noout --relaxng /usr/share/YaST2/schema/autoyast/rng/profile.rng %s
+    # validate: /usr/bin/xmllint --noout --relaxng /usr/share/YaST2/schema/autoyast/rng/profile.rng %s
   with_items: "{{ j2_equipment_groups_list }}"
   when:
     - hostvars[groups[item][0]]['ep_equipment_type'] == 'server'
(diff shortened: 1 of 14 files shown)


oxedions:
Issue detected in kickstart

roles/core/pxe_stack/tasks/client_os/Suse.yml
```diff
@@ -19,7 +19,7 @@
     mode: 0644
     owner: "{{ pxe_stack_apache_user }}"
     group: "{{ pxe_stack_apache_group }}"
-    validate: /usr/bin/xmllint --noout --relaxng /usr/share/YaST2/schema/autoyast/rng/profile.rng %s
+    # validate: /usr/bin/xmllint --noout --relaxng /usr/share/YaST2/schema/autoyast/rng/profile.rng %s
```
neilmunday, line 22: Why has this validation check been commented out?

roles/core/pxe_stack/tasks/client_os/Suse.yml
```diff
@@ -35,7 +35,7 @@
...
```
neilmunday, line 38: And this one?

roles/core/pxe_stack/tasks/client_os/Suse.yml
oxedions: Good question 🙂 
Issue is, these files do not exist on other distributions than opensuse. So I need to comment them for now. However, I kept it here commented, in hope I can find a solution in the future.

neilmunday: No worries - that makes sense. If xmllint exists as a package on other distributions then perhaps it could be a package that is installed by the role? Of course, we could also have to provide a copy of `/usr/share/YaST2/schema/autoyast/rng/profile.rng` from SLES/OpenSUSE.

oxedions: That would make sense, you are right.
I will open an issue for that, to do it in a second step.

oxedions: https://github.com/bluebanquise/bluebanquise/issues/688

neilmunday (approved):

PR #692, opened 2022-04-07
Feat: advanced_dhcp_server - allow time_ip list

thiagocardozo:
Enables the usage of a time_ip list same way as is done with dns_ip.

Thanks to N. Capit.


oxedions:
Dear @thiagocardozo 

Thanks a lot for this PR 😊 
Ok for me, auto check failed du to something external to these changes.

Approved and merged !

PR #699, opened 2022-04-16
OpenSUSE / SLES fixes

neilmunday:
**pxe_stack**

- Handle installation of `sudo` package if sudo usage has been chosen
- Only set product and add on repositories if using SLES - not applicable to OpenSUSE

**repositories_client**

- Update SUSE sub task to better handle a repository definition where a dictionary is use. For example:

```
repositories:
  - name: oss
    disable_gpg_check: true
```

With the current code, this definition causes an error. With my modification this is no longer the case.

### DIFF
--- a/roles/core/pxe_stack/readme.rst
+++ b/roles/core/pxe_stack/readme.rst
@@ -173,6 +173,7 @@ To be done
 Changelog
 ^^^^^^^^^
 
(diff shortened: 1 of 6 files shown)


oxedions (commented):
Dear @neilmunday 

Thanks a lot for this PR. I checked the code, all good for me.

Thanks a lot ! 😊 

Merging.

oxedions (approved):

PR #706, opened 2022-05-06
Fix: powerman - error when ep_host_authentication does not contain IPMI credentials

ginomcevoy:
We had an issue with our customized inventory that provides ep_host_authentication. The automatically generated inventory may contain devices that are considered MC of BMCs, and these devices connect via SNMP. In that case, the list of credentials in ep_host_authentication does not contain any IPMI credentials.

This inventory created an error in powerman role: the "first" filter failed when trying to match the IPMI protocol in the entries.

The proposed fix checks if there are any IPMI credentials in ep_host_authentication before adding credentials to credentials_dict dictionary. It also makes sure not to add the BMC device if there are no IPMI credentials for it (this avoids a "more plugs than nodes" error that happens if the number of devices does not match the number of nodes).

This fix should not affect behavior of inventories that only use the IPMI protocol for credentials.

There are other ways that we could deal with this inventory, e.g. we could use ep_equipment_authentication anyway (not necessary for our use case), or configure powerman role to support SNMP (this would be a new feature).

### DIFF
--- a/roles/core/powerman/templates/powerman.conf.j2
+++ b/roles/core/powerman/templates/powerman.conf.j2
@@ -29,25 +29,36 @@ include "/etc/powerman/ipmipower.dev"
       {% else %}{# BMC is defined outside host #}
         {% set credentials_host = hostvars[host]['bmc']['name'] %}{# Define external bmc as credentials source #}
       {% endif %}
-      {# Gather BMC credentials settings #}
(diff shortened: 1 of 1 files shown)


oxedions:
Dear @ginomcevoy 
Thanks a lot for this PR. Understood the issue, and code is ok to me. Merging. 😊

PR #716, opened 2022-05-20
Fix: ssh - Allow keys to sudo user folder

### LINKED ISSUE #715: [BUG] ssh_remote_keys only add keys to root account
BlueBanquise allows being run as a privileged user. 
Some roles now uses the `ansible_env.SUDO_USER` variable to check which user should be impacted with the role.

ssh_remote_keys isn't (yet).

## Possible Fix

in `tasks/main.yml`, replace `user: root` with:
```yaml
- name: authorized_key █ Add authorized key
  ansible.posix.authorized_key:
    user: "{% if ansible_env.SUDO_USER is defined %}/home/{{ ansible_env.SUDO_USER }}{% else %}/root{% endif %}/.ssh"
    key: "{{ item }}"
    exclusive: no
    state: present
  with_items: "{{ authentication_ssh_keys }}"
```

oxedions:
Fix #715 issue
@wlln please check if it is ok for you 😊 
Once merged, 1.6 release.

### DIFF
--- a/roles/core/ssh_remote_keys/vars/main.yml
+++ b/roles/core/ssh_remote_keys/vars/main.yml
@@ -1,2 +1,2 @@
 ---
-ssh_slave_role_version: 1.1.0
+ssh_slave_role_version: 1.1.1
--- a/roles/core/ssh_remote_keys/vars/main.yml
+++ b/roles/core/ssh_remote_keys/vars/main.yml
@@ -1,2 +1,2 @@
 ---
-ssh_slave_role_version: 1.1.1
+ssh_remote_keys_role_version: 1.1.1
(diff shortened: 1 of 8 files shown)


roles/core/ssh_remote_keys/vars/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-ssh_slave_role_version: 1.1.0
+ssh_slave_role_version: 1.1.1
```
neilmunday, line 2: Should this role still have this variable names? Perhaps it should be `ssh_remote_keys_version`?

strus38: +1

strus38 (changes requested):
Should be good to update rile variable name to the role name.

wlln:
Dear @oxedions 

Apart from the point @strus38 and @neilmunday suggest, I'm fine with this PR.

oxedions:
Thanks a lot for the review ! You are right, my bad. Let me fix that 😊

PR #717, opened 2022-05-22
Fix: flush handlers at the end of role repositories_client

sla31:
Hi,

To allow subsequent roles to install packages.

Thanks!


oxedions (approved):
Dear @sla31 

Thanks a lot for this PR !

Yes, it makes sens. Ok for me ! 😊

PR #718, opened 2022-05-22
Improve idempotency of role pxe_stack

sla31:
Hi!

This is only while creating the Ubuntu cloud-init empty "meta-data" file.  The Ansible module "file" 'touch'es the "meta-data" file every time the role is played (modifying the access time of the "meta-data" file).  Suggesting to replace with the Ansible module "copy" with an empty content.

Thanks!

### DIFF
--- a/roles/core/pxe_stack/tasks/client_os/Ubuntu.yml
+++ b/roles/core/pxe_stack/tasks/client_os/Ubuntu.yml
@@ -73,9 +73,9 @@
     - file
 
 - name: file █ Create cloud-init empty meta-data
-  ansible.builtin.file:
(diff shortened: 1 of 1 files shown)


wlln:
Hi @sla31

Without the `force=no` option in the copy module, the `meta-data` file will be emptied if the file exists.
This should not impact the stack though, as `meta-data` is always empty.

sla31:
Hello @wlln 

That's right!  This PR assumes the "meta-data" is exclusively under the control of the "pxe_stack".  If it's modified outside of "pxe_stack", then "force=no" is necessary.

oxedions (approved):
Dear @sla31,

Thanks a lot for this PR ! :blush:

This was indeed an issue. Code is OK for me, merging !

PR #725, opened 2022-06-09
fix: to prepare bootswitch, retrieve from kernel cmd line only the necessary variables

sla31:
Hi!

Some kernel options may contain dots (like `modprobe.blacklist=`, `rd.driver.blacklist=`, `pciehp.pciehp_debug=`, etc.), so we can't evaluate those options as shell variables when preparing the bootswitch.
The idea here is to only retrieve the necessary kernel cmd line options to evaluate them as shell vars, before curl'ing the bootswitch.

Thanks!


oxedions:
Dear @sla31 

Thanks a lot for this PR 🙂 

It seems ok to me, merging.

PR #726, opened 2022-06-09
fix: pxe_stack role — rationalize single and double quote nested usage

sla31:
Hi!

When tweaking the DGX-OS curtin file to configure all network interfaces in DHCP after 1st reboot in the so-called "late commands", fix improper use of single & double quotes (don't use single quotes for both inner quotes and outer quotes).
Here: using single quotes as outer, and double quotes as inner because when need shell variable expansion.
Also simplifying the code a little.

Thanks!


oxedions:
Dear @sla31 

Thanks a lot for this PR. It looks ok to me, and I don't have a DGX to ensure all of this works in curtin environment of DGXos, so I will trust you blind on that one 😊

PR #729, opened 2022-07-05
Fix bootstrap.sh iso copy [rhel-like]

santos-lucas:
fix a wrong variable call for ${REPO_PATH} on the copy iso step for el8 platform.


oxedions (approved):
Dear @santos-lucas 

Thanks a lot for this PR ! 😊 

Merging !

PR #746, opened 2022-08-29
ci: update to ansible-lint v6.5

btravouillon:
The recent release of ansible-lint v6.5 brings some breaking changes.
Pin the version to 6.5.x, so any future update won't break the CI out of
the blue.

Disable jinja[spacing]: there might be an interest to run it later.

Disable name[casing]: this is not compatible with the task naming
convention of this project.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -4,3 +4,5 @@ warn_list:
 verbosity: 1
 skip_list:
   - fqcn-builtins
+  - jinja[spacing]
+  - name[casing]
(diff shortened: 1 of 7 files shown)


oxedions:
Dear @btravouillon 

Thanks a lot for this PR, and apologies for the delay of my answer.

It seems perfect to me.

Note that for BB 2.0, I will remove special characters (extended ascii ones) on tasks name, as it generates issues on too many external tools.

oxedions (approved):

PR #747, opened 2022-10-24
Fix override.j2

santos-lucas:
Fix override configuration file to use same zone as cluster, otherwise de domain will not be searched correctly


oxedions:
Dear @santos-lucas 

Apologies for the delay. Thanks a lot for this fix, this is ok for me.
I will import it into 2.0 branch.

Issue #756, opened 2022-11-28
[FEATURE REQUEST] Add optionnal 'alias' to every interface.


strus38:
[Provide general introduction to the feature request, and why it is relevant to this repository]

- Add optionnal 'alias' to every interface.
This would apply updating the dns_server and hosts roles.

## Context

It is sometime useful to be able to set an alias per interface instead on only 1 globally. This is the cas for exemple for EMC where 2 interfaces are present on 2 different networks, and where we are forced to define 2 hosts currently, 1 per interface.

## Feature

[Describe the solution you'd like. Please be precise.]

This is  pretty simple feature, where an inventory could be defined like:

```
hosts:
    pm0-emc0:
        alias:
        - pm0-emc0
        network_interfaces:
          - interface: enp0s3
            dhcp_client_identifier: 00:50:18
            ip4: 10.0.0.200
            mac: 08:00:27:dc:f8:f6
            network: ice1-1
          - interface: pm0-emc0-maint
            dhcp_client_identifier: 00:50:63
            network: ice1-2
            ip4: 192.168.0.1
            alias: pm0-emc0-maint
```

PR #757, opened 2022-11-28
#756 - Add optional alias on interfaces

strus38:
This is to implement the #756 feature.


strus38:
@johnnykeats and @oxedions - could you please review?

PR #759, opened 2022-12-02
dns_server: skip networks not in dns

aldarrie:


oxedions:
Dear @aldarrie 

My apologies for being so late, I was off few weeks.
Thank you for this PR.
Good catch !

The dns server role is in the process of being migrated into collections. Path of the role will change. Would you accept to change the PR in few days to adapt to the new path ? I will not correct this bug in collections that will be pushed into master branch, so your contribution is kept in the process.

aldarrie:
Hello,

First of all, Happy New Year!

I do not understand very well.
Has the `is_in_dns` variable been replaced by the `dns_server` variable which is used in the line below?
https://github.com/bluebanquise/bluebanquise/blob/a3211c66242c44b0b920bd7bb8c6f26ed8089434/collections/infrastructure/roles/dns_server/vars/main.yml#L31

If so, you can ignore my PR as this is already taken into account in your edits.

oxedions:
Dear @aldarrie 

Indeed. I just don't remember fixing that 🤔 

Ok, seems it was fixed in BB 2.0 dev branch during this commit:
https://github.com/bluebanquise/bluebanquise/commit/90788128395bae5b779f7d38a700264388fea82c

Sorry, I missed that 😐

aldarrie:
Okay. Thanks.

No worries, we'll wait for the master branch to be stable to update.

PR #764, opened 2023-01-19
monitoring/roles/grafana: allow to run as non-root

pigay:
When running role grafana as non-root user within a playbook that uses `become: true`, the download of dashboards is performed by root.

However, following tasks that loop over `with_fileglob` cannot find downloaded files because the fileglob is performed by non-root playbook caller and cannot read download directory.

Adding `become: false` in the download part allows to correctly find the files.

### DIFF
--- a/collections/monitoring/roles/grafana/tasks/dashboards.yml
+++ b/collections/monitoring/roles/grafana/tasks/dashboards.yml
@@ -2,6 +2,7 @@
 - name: dashboards █ Download Opensource Grafana dashboards
   delegate_to: localhost
   run_once: true
+  become: false
   block:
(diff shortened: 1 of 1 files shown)


oxedions:
Dear @pigay,

Thanks a lot for this PR ! 🙂 
Ok for me. Could you also add your name as authors/contributors in https://github.com/bluebanquise/bluebanquise/blob/master/collections/monitoring/galaxy.yml, and increment in the same file version from 2.0.0 to 2.0.1 since it is a bug fix ? That would be perfect.

pigay:
Done ;)

oxedions:
Thats perfect ! 😊 
Many thanks for this improvement ! I will also merge your improvement made on community asap 😉

PR #783, opened 2023-02-13
[Fix] - Apply fix to correct rsyslog role

thiagocardozo:
In #771, part of the fix was applied in the old "legacy" version of the role instead of the desire new in collections.


thiagocardozo:
Should we also rollback changes in the legacy role?

oxedions:
Dear @thiagocardozo 
You are right, I missed that.
Thank you for this fix 😊 
Merging!

oxedions:
Sorry @thiagocardozo , I missed your second message.
I will soon delete the legacy folder, it was only here for temporary convenience 🙂

thiagocardozo:
I still need one more change in this role. It seems that rsyslog might still write to rotated logs after the service is restarted.
This [behavior](https://access.redhat.com/discussions/4500431) was observed in some places.

PR #786, opened 2023-02-15
infrastructure/dns_server: find correct default resolution network in reverse zone

pigay:
## Describe your changes

Reverse zone definition templates used j2_node_main_resolution_network to select main interface and write short name, but variable `item` is a network address and not a name.

We propose to check against host_j2_node_main_resolution_address prefix address instead

- [X] Document introduced changes / variables in role README.md if any

### DIFF
--- a/collections/infrastructure/roles/dns_server/README.md
+++ b/collections/infrastructure/roles/dns_server/README.md
@@ -52,6 +52,7 @@ This will cause `/var/named/override` to be generated.
 
 ## Changelog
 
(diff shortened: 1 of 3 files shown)


oxedions:
Dear @pigay 

Indeed ! Thanks a lot for this fix ! I introduced this bug during 1.6 version, while upgrading the role.
I will need to focus a bit on this role once 2.0 version is stable, as it is really slow to operate, and there are possible improvements to speedup templates rendering for large clusters 😊 

Ok for me, merging !

PR #789, opened 2023-02-17
infrastructure/firewall: enforce firewalld AllowZoneDrifting

pigay:
## Describe your changes

Firewalld is configured with `AllowZoneDrifting=yes` by default on RedHat systems, which seems to be a security risk.

We propose to enforce configuration to `no` by default (based on an ansible variable to allow configuratbility)

- [X] Document introduced changes / variables in role README.md if any

### DIFF
--- a/collections/infrastructure/roles/firewall/README.md
+++ b/collections/infrastructure/roles/firewall/README.md
@@ -111,14 +111,22 @@ Optional inventory vars:
     * icmp_block_inversion (bool)
     * masquerade           (bool)
 
+Optional role vars:
+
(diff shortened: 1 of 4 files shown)


oxedions:
Dear @pigay 

Thank you for this PR ! 😊 
Could you just adapt the new variable name to start with `firewall_` prefix, which is the role name ?
Even if it sounds a bit too much, it could be `firewall_firewalld_allow_zone_drifting`, that is ok. Everything else is ok to me.

pigay:
Sure, done ;)

oxedions:
Thanks a lot ! 😊 
Ok for me, merging !

PR #796, opened 2023-03-09
infrastructure/firewall: ensure nics are in the correct fw zone

pigay:
## Describe your changes

For nodes with multiple NICs, firewalld role didn't set each one in its own firewall zone.


oxedions:
Dear @pigay 

Yes, I agree with this change. It was missing in the role.
Thank you for this PR 🙂 

Ok for me !

PR #827, opened 2023-06-30
Bootstrap dependencies error

hmeNLE:
Hello,

I suggest a minor fix that adds a dependency to the automatic configuration of the ansible and bluebanquise environment.

patch works on alma 9.2

Best

the error was as follows: 

```
ERROR! Failed to resolve the requested dependencies map. Could not satisfy the following requirements:
* bluebanquise.commons:* (dependency of bluebanquise.infrastructure:2.1.2)
```


oxedions (approved):
Dear @hmeNLE 
Many thanks for this fix ! 😊 
Indeed, I forgot to update bootstrap while adding comment collection.
I also notice that some tests are failing but its not related to this PR.
Ok for me, merging !

PR #832, opened 2023-08-11
Fix prometheus service item

ginomcevoy:
## Describe your changes
Added a condition in the prometheus role, prevents error when an exporter has no service parameter.


oxedions (approved):
Dear @ginomcevoy 

Thank you very much for this PR and the fix 😊 
Indeed, this is a bug.

Ok for me, merging !

PR #835, opened 2023-08-29
Fix prometheus restart services

ginomcevoy:
## Describe your changes
Restart each service managed by the prometheus role when its systemd file changes. This is relevant when the launch parameters change: the services were not being updated automatically with new launch parameters.

Also changed the order of the handlers in handlers/main.yml, the handler for the systemd daemon-reload must go before the service restarts. FYI, the handlers are executed in the order they are defined, not in the order they are listed in the notify statement: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#handlers

### DIFF
--- a/collections/monitoring/roles/prometheus/README.md
+++ b/collections/monitoring/roles/prometheus/README.md
@@ -581,6 +581,7 @@ prometheus_server_prometheus_launch_parameters: |
 
 ## Changelog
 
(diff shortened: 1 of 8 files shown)


oxedions (approved):
Dear @ginomcevoy 

Many thanks for this PR and fix ! 😊 
Looks good to me, you are right. Merging !

Issue #923, opened 2024-04-12
[FEATURE REQUEST] Infrastructure/repositories role: set a default network


jpm38:
This feature request is to be able to set a default network when configuring repositories client and so be able to override this value

jpm38:
PR has been merged

Issue #920, opened 2024-04-03
[BUG] Repositories role should handle RedHat os/BaseOS and os/AppStream repos and description should be set to name


jpm38:
Two regressions have been introduced in repositories role:

1)

the following repositories inventory no longer works:
- {name: os, gpgcheck: 1, gpgkey: "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release"}

This is suppose to generate for RedHat, BaseOS and AppStream repositories client

2)

- name: yum_repository <|> Setting repositories
  ansible.builtin.yum_repository:
    name: "{{ item.name | default(item) }}"
    description: "{{ item.description | default(item) }} gen by Ansible"
    baseurl: "{{ item.baseurl | default(repositories_j2_simple_url + (item.name | default(item))) }}/"
 
The module yum_repository.py overrides name with description and so the result in /etc/yum.repos.d is wrong.

In the previous bluebanquise version, the description was :
description: "{{ item.name | default(item) }} gen by Ansible"

jpm38:
PR has been merged.

PR #925, opened 2024-04-12
Set default network and handle RedHat os

jpm38:
## Describe your changes

This PR is to fix the two tickets hereafter.

#920 and #923

### DIFF
--- a/collections/infrastructure/roles/pxe_stack/README.md
+++ b/collections/infrastructure/roles/pxe_stack/README.md
@@ -717,6 +717,7 @@ Note that using an home folder into /home for the bluebanquise sudo user can be 
 
 ## Changelog
 
+* 1.16.5: Several fixes to bluebanquise-diskless (version 2.0.7). Giacomo Mc Evoy <gino.mcevoy@gmail.com>
 * 1.16.4: Ensure bluebanquise-bootset is run as root/sudo. <patrick.begou@univ-grenoble-alpes.fr>
 * 1.16.3: Fix bug in kickstart (reported by jpm38). Benoit Leveugle <benoit.leveugle@gmail.com>
 * 1.16.2: Fix owner and group for bluebanquise .ssh directory. <patrick.begou@univ-grenoble-alpes.fr>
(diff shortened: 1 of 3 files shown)


collections/infrastructure/roles/repositories/tasks/RedHat/main.yml
```diff
@@ -20,7 +20,7 @@
 - name: yum_repository <|> Setting repositories
   ansible.builtin.yum_repository:
     name: "{{ item.name | default(item) }}"
-    description: "{{ item.description | default(item) }} gen by Ansible"
+    description: "{{ item.name | default(item) }} gen by Ansible"
```
oxedions, line 23: Can you change that to:

`"{{ item.description | default(item.name, true) | default(item, true) }} gen by Ansible"`

?
Idea is to allow description, and fallback on name then on item if not set or None

collections/infrastructure/roles/repositories/vars/main.yml
```diff
@@ -1,4 +1,4 @@
...
+repositories_j2_simple_url: "http://{{ networks[repositories_client_network]['services']['repositories4'][0]['ip4'] | default(networks[repositories_client_network]['services']['repositories'][0]['ip4'], true) | default(networks[repositories_client_network]['services_ip'], true) | default('', true) }}/repositories/{{ os_operating_system['repositories_environment'] | default('', true) }}/{{ os_operating_system['distribution'] | default('', true) }}/{{ os_operating_system['distribution_version'] | default(os_operating_system['distribution_major_version']) | default('', true)}}/{{ ansible_architecture }}/"
```
oxedions, line 4: Can you rename this variable to `repositories_network` instead ? The role has been renamed.

Also, I would prefer to use default here, instead of setting a j2 in defaults, to avoid losing synchronisation between roles.

Could set

`repositories_client_network:`

In defaults, and here:

`networks[repositories_client_network | default(j2_node_main_network, true)]`

?

Both way works, just to ensure all roles work the same way :slightly_smiling_face: 

Can you also add a small ntoe in the README file about this new variable ? Just 2 lines, to expose it.

oxedions (changes requested):
Thanks a lot for this PR, and very sorry for the delay :blush: 

Few adjustments, but ok for the concept.

collections/infrastructure/roles/repositories/vars/main.yml
jpm38: Thanks for the review. Will apply the requested changes

oxedions:
Note for me : other variables in the repository role do not match naming !

jpm38:
Reopen

jpm38:
Open the PR

jpm38:
Reopen

PR #926, opened 2024-04-14
infrastructure/ssh_client role: set a default network

jpm38:
- [x ] Increment role version in vars/main.yml (a.b.c: +1 to b for a feature added, +1 to c for a bug fix)
- [x ] Update changelog inside role README.md
- [ ] If not present, please also add your name in the related collection authors list in galaxy.yml

### DIFF
--- a/collections/infrastructure/roles/ssh_client/README.md
+++ b/collections/infrastructure/roles/ssh_client/README.md
@@ -89,8 +89,14 @@ To achieve that, the following variables are available:
 * ssh_client_global_loglevel
 * ssh_client_global_stricthostkeychecking
 * ssh_client_global_userknownhostsfile
+* ssh_client_global_network
 
-Which are evaluated at global level, and:
+Which are evaluated at global level
+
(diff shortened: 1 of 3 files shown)


oxedions (changes requested):
Dear @jpm38 

Thank you for this PR :blush: 

Since the set is already using a piping and default, would you accept to change this by adding another default ?
Also since your usage is global, could you rename the variable to match other global values ?

So something like :

`set host_to_be_used_network = ssh_client_global_network  | default(host_vars['j2_icebergs_main_network_dict'][j2_current_iceberg], true) | default(none, true)`

Also, please add a small note in the README.md in the advanced usage part so that everyone is aware of this variable :slightly_smiling_face:

oxedions (approved):
Dear @jpm38 

Apologies for the delay of my answer, its kids holidays, so I have dad's duties that uses all my time.

This is perfect to me, thank you very much for the change ! :blush:

PR #937, opened 2024-05-10
BB-35

oxedions:
Fix multiple reported issues and add requested features.


sgaosdgr:
Thank you for quick response.

At line 38 of client.conf.j2, "fwdRule{{ loop.count }}" does not work.

```
38  $ActionQueueFileName fwdRule{{ loop.count }} # unique name prefix for spool files
```

I had to remove {{ loop.count }} to get it working. Can you double check?

oxedions:
@sgaosdgr 

It should be better now. I don't understand why I used loop.counter instead of loop.index. Probably because I switch languages all the time during days between job and opensource, so I end up mixing them :blush:

PR #940, opened 2024-05-17
Update kickstart.cfg.j2 for RedHat to assign  /var/lib/bluebanquise/.ssh to bluebanquise user

patrick-begou:
**Update kickstart.cfg.j2 for RedHat to assign  /var/lib/bluebanquise/.ssh to bluebanquise user**

When deploying RedHat based OS on nodes and creating ssh keys from the kickstart.cfg, files in bluebanquise home directory belong to root and not to bluebanquise user. A chown is added.

## Describe your changes
Added a chown command in the RedHat template file to solve the problem.

- [ ] Increment role version in vars/main.yml (a.b.c: +1 to b for a feature added, +1 to c for a bug fix)
- [ ] Update changelog inside role README.md
- [ ] If not present, please also add your name in the related collection authors list in galaxy.yml

### DIFF
--- a/collections/infrastructure/galaxy.yml
+++ b/collections/infrastructure/galaxy.yml
@@ -4,7 +4,7 @@ namespace: bluebanquise
 name: infrastructure
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 3.10.1
+version: 3.10.2
 
 # The path to the Markdown (.md) readme file. This path is relative to the root of the collection
 readme: README.md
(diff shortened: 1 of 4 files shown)


patrick-begou:
I had always used "." as a separator between user and group in chown command with RedHat based OS but you are right, the manual says it should be ":".

oxedions:
Dear @patrick-begou 

Many thanks for this fix. It seems ok to me now :blush: 

Merging !

Issue #942, opened 2024-05-24
BB-36 - [BUG] Slurmdbd.conf.j2 template not working with a remote dedicated mariadb sever


sgaosdgr:
Template https://github.com/bluebanquise/bluebanquise/blob/master/collections/hpc/roles/slurm/templates/slurmdbd.conf.j2  does not work well when slurm database is on another host from where slurmdbd is running. 

## Context

For slurm, `DbdHost` refers to the machine where slurmbdb is running, `StorageHost` the machine where mariadb server is running.  `DbdHost` and `StorageHost` can be the same machine when running local mariadb/mysql database server. But when mariadb/mysql is running on another machine, `DbdHost` and `StorageHost` would be different. 

To be consistent, it's good to always use `DbdHost` and `StorageHost` variables.

Here is the updated version that works for both cases:
```
#jinja2: lstrip_blocks: True
#### Blue Banquise file ####
## {{ansible_managed}}

# Documentation:
# https://slurm.schedmd.com/slurmdbd.conf.html

## Authentication
AuthInfo=/var/run/munge/munge.socket.2
AuthType=auth/munge
SlurmUser=slurm

{% if slurm_enable_local_mysql is defined and slurm_enable_local_mysql %}
## Mariadb server
DbdHost={{ slurm_accounting_mysql_login_host | default('localhost', true) }}
{% else %}
# SlurmDdb info
DbdAddr={{ slurm_contoller_ip_address | default(ansible_default_ipv4.address) }}
DbdHost={{ slurm_controller_hostname | default(ansible_hostname) }}
DbdPort={{ slurm_dbd_port | default('6819') }}
MessageTimeout=10
DefaultQOS=normal
PluginDir=/usr/lib64/slurm
TrackWCKey=yes
{% endif %}

## Log and pid
LogFile={{ slurm_slurmdbdlogfile }}
DebugLevel={{ slurm_slurmdbddebug }}
PidFile=/var/run/slurmdbd.pid

## Storage
StorageType=accounting_storage/mysql
{% if slurm_enable_local_mysql is defined and slurm_enable_local_mysql %}
StorageUser={{ slurm_accounting_storage_user }}
StoragePass={{ slurm_accounting_storage_pass }}
{% else %}
StorageHost={{ slurm_accounting_mysql_login_host }}
StoragePort={{ slurm_accounting_mysql_login_port }}
StorageUser={{ slurm_accounting_mysql_login_user }}
[...]

Issue #941, opened 2024-05-23
BB-36 - [BUG] Need handler for scontrol reconfigure after updates to slurm.conf


vedmonds:
When slurm.conf is updated, the slurm controller needs to issue `scontrol reconfigure`

## Context

When `scontrol reconfigure` is issued it:
1. verifies the new slurm.conf is valid
2. alerts compute nodes that there is a new version of slurm.conf (critical for when running in configless mode)

## Process

1. Update slurm.conf to include new hosts
2. Install slurmd on computes in configless mode
3. `systemctl status slurmd` on new computes will fail

## Expected result

Expect slurmd service to be running properly on new computes

## Current result

```
    Unable to start service slurmd: Job for slurmd.service failed because the control process exited with error code.
    See "systemctl status slurmd.service" and "journalctl -xeu slurmd.service" for details.
```

## Possible Fix

`collections/hpc/roles/slurm/tasks/controller.yml`
```
- name: "template <|> Generate configuration files in {{ slurm_home_path }}"
  ansible.builtin.template:
    src: "{{ item }}.j2"
    dest: "{{ slurm_home_path }}/{{ item }}"
    owner: slurm
    group: slurm
    mode: 0644
  tags:
    - template
  loop:
    - slurm.conf
    - cgroup.conf
  notify: Run scontrol reconfigure
```

`collections/hpc/roles/slurm/handlers/main.yml`
```
- name: Run scontrol reconfigure
  ansible.builtin.command:
    cmd: scontrol reconfigure
  changed_when: true
```
## Additional data

[If relevant, include a screenshot, links, etc.]

oxedions:
Dear @vedmonds 

Apologies for the delay.

Yes, your issue make sens. I will fix that asap :blush:

oxedions:
Dear @vedmonds 
I added a fix in https://github.com/bluebanquise/bluebanquise/pull/944, can you check its ok for you ?

PR #944, opened 2024-06-10
BB-36

oxedions:
Fix slurm issues #942 and #941


sgaosdgr:
Thank you for the updates. I will run some tests to confirm.

oxedions:
Dear @sgaosdgr 

I upgraded packages to 23.11.8. Please be careful when migrating slurm database, its always best to backup it before a major upgrade.

I also removed tcp debug in ipxe, since its not needed anymore.

Note that if you are using ubuntu or debian, I switched to DEB native packages (slurm DEB support was introduced in 23.11), I will update the slurm role soon to match new packages names. This has no impacts on RHEL.

Last note, are you interested in slurmrestd ? Its HTTP REST API. I can easily add it in packages builder steps.

PR #973, opened 2024-09-30
add support for multi-rail on LNET configuration

santos-lucas:
## Describe your changes

Just add support for multi-rail configuration on the LNET, if there is more than one interface into the same network.


oxedions (changes requested):
Dear @santos-lucas 

Thank you very much for this PR.
Code seems ok to me.

Can you just fix the version missmatch between readme and vars ? You set version  as 1.2.0 in readme while it should be 1.1.0.
Also, if possible, can you update the version inside collections/file_systems/galaxy.yml file ? So that everyone using this collection can easily update it :blush:

PR #980, opened 2024-10-13
Add custom_packages role

santos-lucas:
## Describe your changes

Creation of custom_packages role, to manage the installation of packages that are not already handled by other roles.

### DIFF
--- a/collections/infrastructure/roles/custom_packages/README.md
+++ b/collections/infrastructure/roles/custom_packages/README.md
@@ -15,6 +15,19 @@ I recoomend you to have on single file on the "all" group ( $HOME/bluebanquise/i
 to manage all the custom packages in the system, as the role will already filter the instalation by groups
 and normally will be more visible to have this on a single file
 
+Example of "custom_packages.yml" file:
+
+```yaml
+custom_packages:
+  mg_computes:
+    - openmpi
+  mg_logins:
+    - vim
+    - '@Development Tools'
+  equipment_typeGPU:
+    - htop
+```
+
 ## Changelog
 
 * 1.0.0: Role creation. Lucas Santos <lucassouzasantos@gmail.com>
--- a/collections/infrastructure/roles/custom_packages/README.md
+++ b/collections/infrastructure/roles/custom_packages/README.md
@@ -1,0 +1,20 @@
+# Custom Packages
+
+## Description
+
+This Role simply allows to install custom packages on the system that 
+are not provideded to other bluebanquise roles.
+That can be usefull to keep all the customizations relating OS packages
+synced with BlueBanquise so you dont need to have extra steps in 
+configuring nodes.
+
+## Instructions
+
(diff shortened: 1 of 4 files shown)


collections/infrastructure/roles/custom_packages/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+- name: "Install sequana packages"
+  yum:
+    name: "{{sequana_packages[item]}}"
```
hmeScaler, line 3: maybe 
```suggestion
    name: "{{custom_packages[item]}}"
```

santos-lucas:
You are right @hmeScaler , many thanks to pointing that out.

I guess I got the fisrt version that I did for some sequana cluster, just fixed that + the other mentions with "sequana"

collections/infrastructure/roles/custom_packages/README.md
```diff
@@ -0,0 +1,33 @@
+# Custom Packages
+
+## Description
+
+This Role simply allows to install custom packages on the system that 
+are not provideded to other bluebanquise roles.
```
oxedions, line 6: provideded -> typo

collections/infrastructure/roles/custom_packages/README.md
```diff
@@ -0,0 +1,33 @@
...
+That can be usefull to keep all the customizations relating OS packages
+synced with BlueBanquise so you dont need to have extra steps in 
+configuring nodes.
+
+## Instructions
+
+This role is very simple and works based on ansible groups.
+I recoomend you to have on single file on the "all" group ( $HOME/bluebanquise/inventory/group_vars/all/addons/custom_packages.yml)
```
oxedions, line 14: recoomend -> typo

oxedions (commented):
Dear @santos-lucas 

Many thanks for this new role ! :blush: 
It is very close to https://github.com/bluebanquise/community/tree/main/roles/generic_psf

I was wondering: why filtering  by groups inside the role ? I mean by that: isn't it better to let users set variable per group in the inventory, and just load the hostvars to get node packages to install ?

santos-lucas:
Hello @oxedions ,

I currently do it like this, because I use this in a couple cluster with hetereogeneous compute nodes, so it makes easier to look into the custom_packages configuration for the whole cluster, otherwise I would need to navigate between the group_vars directories and makes even harder to compare something if needed.

It is kinda of the same thing as we have for the host_vars in bluebanquise, we use everything on a single file instead of using the host_vars directory. For this single pruporse of the "custom_packages" it makes life a little easier, even more for sysadmins that work on systems with BlueBanquise but are not too experienced with it.

I see that it seems a lttle weird with the BB structure, but that is why I created it like this.

oxedions:
Dear @santos-lucas 

Thank you for this honest feedback.
This is fair, ok for me! :blush:

oxedions (approved):

PR #986, opened 2024-11-06
NFS - Fixed skipping server if using IP; Add parameter to run role on server group

thiagocardozo:
## Describe your changes
Role was skipping host if using IP because the check used hostname strictly.
Added a server_group parameter to be used for cases like HA, where clients would want to point to a VIP.


collections/file_systems/roles/nfs/tasks/server.yml
```diff
@@ -61,7 +61,9 @@
     path: /etc/exports
     regexp: ' # export id: {{ item.export }}$'
     line: "{{ item.export }} {% if item.network is defined and item.network is not none and networks[item.network] is defined and networks[item.network]['subnet'] is defined and networks[item.network]['subnet'] is not none and networks[item.network]['prefix'] is defined and networks[item.network]['prefix'] is not none %}{{networks[item.network]['subnet']}}/{{networks[item.network]['prefix']}}{% else %}*{% endif %}({{ item.export_options | default('rw,sync,root_squash', true) }}) # export id: {{ item.export }}"
-  when: inventory_hostname == item.server
+  when: (inventory_hostname == item.server) or
+        (item.server in ansible_all_ipv4_addresses) or
+        (inventory_hostname in groups[item.server_group])
```
oxedions, line 66: Dear @thiagocardozo 

Thank you very much and apologies for the delay of my answer :blush: 

I understand the need. I am however afraid of this groups check, because if group do not exist, it might crash, while this group should remain optional. Could you mark this value as optional in the README and also add a check here to ensure the group exists before quering it ?
Also, please update the collection version.

oxedions (changes requested):

PR #987, opened 2024-11-07
FIX: pxe_stack for RH causing duplicated PermitRootLogin entry in sshd_config

thiagocardozo:
## Describe your changes
The sed command present in the kickstart template has a too inclusive regexp,
resulting in the substitution of a PermitRootLogin parameter description in a documentation block,
which leads to 2 repetitive entries.
So we narrow it down to the line starting with "#PermitRootLogin", which is the actual commented out option.

### DIFF
--- a/collections/infrastructure/galaxy.yml
+++ b/collections/infrastructure/galaxy.yml
@@ -4,7 +4,7 @@ namespace: bluebanquise
 name: infrastructure
 
 # The version of the collection. Must be compatible with semantic versioning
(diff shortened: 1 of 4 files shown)


oxedions (changes requested):
Dear @thiagocardozo 

Many thanks for this fix ! :blush: 
This is ok for me.
Could you just please just update the collection version ? (x.y.z, here z since its a fix).

oxedions (approved):

PR #989, opened 2024-11-13
Fix: HA pcs - Task doesn`t find file on running node

thiagocardozo:
When the task to load a cib configuration file runs in the secondary mngt node, we get the following error:

`stderr: 'Error: unable to parse new cib: [Errno 2] No such file or directory: ''/root/ha_config.ansible.xml''`

As it happens the task that stores the file previously runs only on mngt1.
This PR changes that in that every node will create the cib file locally.

### DIFF
--- a/collections/high_availability/galaxy.yml
+++ b/collections/high_availability/galaxy.yml
@@ -4,7 +4,7 @@ namespace: bluebanquise
 name: high_availability
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 3.1.1
+version: 3.1.2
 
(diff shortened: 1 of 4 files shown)


oxedions (changes requested):
Dear @thiagocardozo 

Nice catch ! Thanks a lot for this fix :blush: 

Could you update the collection version too here ? Code is ok to me.

oxedions (approved):
Dear @thiagocardozo 

Thank you very much for this PR ! :blush: 
Ok for me, merging !

PR #990, opened 2024-11-13
Fix/Feat: Grafana - fix uid/gid conflict; Allow custom firewall port

thiagocardozo:
* Grafana user/group ids are usually created at package installation, so let it empty as default to avoid conflicts
* Grafana user changed to system type
* Set log owner by name as uid may vary with change above
* Allow setting custom port in firewalld

### DIFF
--- a/collections/monitoring/roles/grafana/README.md
+++ b/collections/monitoring/roles/grafana/README.md
@@ -151,6 +151,7 @@ Note: if you try to add dashboards, the role will alwats at checking if the data
 
 ## Changelog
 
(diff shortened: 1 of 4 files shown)


oxedions (changes requested):
Dear @thiagocardozo 

Thank you for this PR.

Few questions:

* Why removing the home entry in the user creation ?
* Can you check line 44 in install.yml file, it generates a static check error.
* Can you update the collection version ?

Thanks :blush:

thiagocardozo:
The change is to turn grafana user to the system category, in which you don`t login and therefore don`t have a home.
Similar to apache and postgres users.

oxedions (approved):
Dear @thiagocardozo 

Ok, make sens to me. Thank you !
Merging !

PR #992, opened 2024-12-03
Feat: podman -  Update registries template/use handlers/replace deprecated config

thiagocardozo:
* The [registries.conf](https://github.com/bluebanquise/bluebanquise/blob/master/collections/containers/roles/podman/templates/etc/containers/registries.conf.j2) V1 format[ was deprecated](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md) and the format is not rendering correctly in RHEL 8.6:

> Jun 20 13:18:27 management1 systemd[1]: Starting registry.service...
Jun 20 13:18:27 management1 podman[102046]: Error: loading registries configuration "/etc/containers/registries.conf": toml: line 18 (last key "regi>
Jun 20 13:18:27 management1 systemd[1]: registry.service: Main process exited, code=exited, status=125/n/a

Lets switch to V2.

* Use handlers to restart services when config files are modified.

* libpod.conf is [deprecated and replaced](https://podman.io/blogs/2020/12/11/remove-varlink-libpod-conf-notice) by containers.conf


giacomo-mc-evoy:
A couple of points:
- @thiagocardozo, the PR includes "podman_registry_enable_encryption" variable, this should not be present here. This line could be removed I believe
```
{% if insecure_registry != private_local_registry or not podman_registry_enable_encryption %}
```
- I believe we should increase the minor version of the role/collection, currently only the fix version of the collection is increased, @oxedions  can you please confirm? 
- We got `OSError: [Errno 28] No space left on device` on the check of EL8, can this be safely ignored?

oxedions (approved):
Dear @thiagocardozo 

Impressive PR ! Many thanks :blush: 

Ok for me !

PR #993, opened 2024-12-04
Feat: dns_server - round robin load balance

thiagocardozo:
The idea is to create aliases that point to the same IP without also being inserted in /etc/hosts.

So this entry:

```
fn_management:
  hosts:
    mgt1
      network_interfaces:
        - interface: eth0
          ip4: 10.0.0.10
          network: ice1-1
          dns_alias:
            - www
    mgt2
      network_interfaces:
        - interface: eth0
          ip4: 10.0.0.20
          network: ice1-1
          dns_alias:
            - www
```

Will create a `foward.zone` file with:

```
  www IN A 10.0.0.10
  www IN A 10.0.0.20
```

A lookup to the webserver should return:

```
www.bluebanquise.example.com has address 10.0.0.10
www.bluebanquise.example.com has address 10.0.0.20
```

Ref:
https://access.redhat.com/solutions/54311


ginomcevoy:
Hello @thiagocardozo , can you please resolve the conflict? Probably need to check the role version.
After that, feel free to merge, tested OK

thiagocardozo:
Reopened at https://github.com/bluebanquise/bluebanquise/pull/1061

PR #1011, opened 2025-03-01
Fix: powerman - Expand powerman nodeset for ipmipower

thiagocardozo:
ipimipower isn't expanding nodesets with multiple intervals properly, ex:

> ipmipower -h head_[0-1]-tail_[0,1]
head_0-tail_[0
head_1-tail_[0

A workaround is to do a prior expansion in powerman config:

> ipmipower -h head_0-tail_0,head_0-tail_1,head_1-tail_0,head_1-tail_1
head_0-tail_0
head_0-tail_1
head_1-tail_0
head_1-tail_1


thiagocardozo:
Needs a better solution for this as expanding huge nodesets by default will be problematic.

PR #1023, opened 2025-03-16
Feat: comnan - Added ssh settings for conman

thiagocardozo:
When using conman with redfish(ssh) a known_hosts file is create for user conman.
This leads to issues when performing changes in HW, which won't update the file.


thiagocardozo:
pxe_stack changes are coming from master branch.

PR #1024, opened 2025-03-16
Feat: comnan - Added ssh settings for conman

thiagocardozo:
When using conman with redfish(ssh) a known_hosts file is create for user conman.
This leads to issues when performing changes in HW, which won't update the file.


oxedions (dismissed):
Dear @thiagocardozo 

It makes sens, thank you. I didn't had a RedFish compatible hardware to test it, so with this fix it will be functional, many many thanks :blush: 

Ok for me. Can you just fix the small conflict ? I don't have the rights to do it on your fork.

oxedions:
Dear @thiagocardozo 
This is ok for me, but could just do not edit galaxy.yml version ?
I will try now to only update it with BB versions releases :slightly_smiling_face:

Issue #1008, opened 2025-02-14
[BUG] bb_domain_name not considered by bluebanquise.infrastructure.dns_client


giacomo-mc-evoy:
bb_domain_name not considered by dns_client role by default, bb_domain_name is only used if dns_client_domain_name has been set to an empty string. This means that the search in /etc/resolv.conf will be set to "cluster.local" unless dns_client_domain_name  is overridden, regardless of the value of bb_domain_name.

## Context

Using dns_client role produces a (maybe) unexpected result in /etc/resolv.conf if dns_client_domain_name  is not defined in inventory.

## Process

1. Define bb_domain_name in inventory, e.g. "mydomain.com"
2. Do not set dns_client_domain_name   in inventory
3. Run the dns_client role.

## Expected result

I believe that the expected result is for bb_domain_name to be used, e.g. in /etc/resolv.conf (assuming bb_domain_name is set to "mydomain.com"

```
#### Blue Banquise file ####
## Ansible managed file, do not edit

search mydomain.com
nameserver 10.10.0.1
```

## Current result

The default "cluster.local" is used, e.g. in /etc/resolv.conf

```
#### Blue Banquise file ####
## Ansible managed file, do not edit

search cluster.local
nameserver 10.10.0.1
```

## Possible Fix

Suggestion is to remove "dns_client_domain_name" from defaults/main.yml, or set it to '', for example
bluebanquise/infrastructure/roles/dns_client/defaults/main.yml

```
---
dns_client_domain_name: ''
```

The effect is that the role will use bb_domain_name if dns_client_domain_name remains unset in inventory. The dns_client_domain_name variable will still have precedence as intended.

PR #1025, opened 2025-03-16
Fix: dns_client - DNS client default domain overriding

thiagocardozo:
dns_client_domain_name has a default value set, and since it takes precedence over bb_domain_name, which is global,
dns_client always sets resolv.conf to the default value.

The template also has a default value hardcoded directly, so an empty variable will not fail:

`search {{ dns_client_domain_name | default(bb_domain_name, true) | default('cluster.local', true)}} `

### DIFF
--- a/collections/infrastructure/galaxy.yml
+++ b/collections/infrastructure/galaxy.yml
@@ -5,7 +5,7 @@ name: infrastructure
 
 # The version of the collection. Must be compatible with semantic versioning
 
-version: 3.2.2
+version: 3.2.3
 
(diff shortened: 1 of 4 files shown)


ginomcevoy:
FYI this should resolve [issue 1008](https://github.com/bluebanquise/bluebanquise/issues/1008)

oxedions (approved):
Dear @thiagocardozo 

Ok for me ! Thank you for this PR.

Please note that this role should be now deprecated, and only be used for debugging purposes. DNS client is managed by NetworkManager or SystemdNetworkd, this file is automatically edited by these tools and so a template should not overwrite it.

PR #1039, opened 2025-03-18
Amd gpu

santos-lucas:
## Describe your changes

1) Add support for amd GPUs ( tested only on MI300A ).
2) Make sure gather facts is executed on the beggining of the role. This is needed because when you have an kernel upgrade on the same playbook, ansible_kernel "fact" end up being the one on the begining of the playbook, what leads to problems when instaling kernel-headers, kernel-devel and so on.

- [x] Increment collection version in the galaxy.yml file of the infrastructure collection (a.b.c: +1 to b for a feature added, +1 to c for a bug fix). If not present, please also add your name in the related collection authors list in galaxy.yml
- [x] Update CHANGELOG file at repository root.


oxedions (approved):
Dear @santos-lucas 

Many thanks for this enhancement !
I see more and more AMD GPUs on the market, so this is a real plus to support them ! :blush: 

Ok for me, merging !

PR #1054, opened 2025-03-22
Fix: grafana -  service start in BB3, user creation error

thiagocardozo:
- Replaced start_services by bb_start_services
- Fixed user creation check error(string to int conversion)


ginomcevoy:
All good for me, tested default uid/gid and custom uid/gid.
Just resolve the conflict and feel free to merge.

PR #1058, opened 2025-03-26
Limit some roles in infrastructure collection

thiagocardozo:
Removed the following files from infrastructure collection:

plugins/modules/networkd
roles/singularity
roles/gpu
roles/interconnect
roles/nhc
roles/slurm
roles/google_authenticator


ginomcevoy:
Just what we needed, thanks!

PR #1060, opened 2025-03-28
Feat: auditd - Log root commands

thiagocardozo:
Added option to allow logging root commands.


oxedions (approved):
Dear @thiagocardozo 

Ok for me ! :blush: 
Thank you !

PR #1062, opened 2025-03-28
Feat: podman - podman registry https support

thiagocardozo:
Adds option to enable encrypted registries.


ginomcevoy:
Tested OK with local CA (self-signed) that generated an X.509 certificate for the registry.

PR #1064, opened 2025-04-02
feat: pxe_stack allow selection of shutdown action other than reboot & allow to keep osdeploy boot

jilkarnas:
## Describe your changes

- added variable `pxe_stack_post_install_action` to be able to poweroff of halt instead of reboot (default). Affect Redhat, Ubuntu and Debian.(https://www.debian.org/releases/stable/example-preseed.txt) but not tested.
- added variable `pxe_stack_post_install_boot_to_disk: true` to be able to keep booting with osdeploy (with false). Affect all OS, but not CloneZilla, which has one config file for all nodes if I understood correctly so I avoided changing it.

Original use case is for the deployment of RHEL 9.5 on a Proxmox VM. I only use BB to manage PXE stack but boot order is managed elsewhere. For the change of boot order to be applied on Proxmox VM, poweroff is needed (not reboot).

Ubuntu and Debian adapations (not tested) of first variable are based on [official preseed example] (the d-i finish-install/reboot_in_progress note only deactivate "pop-up" message, reboot being the default action).

Adaptations of second variable for other os (Ubuntu, Debian, Suse) was straight-forward.

- [x] Make sure your name is present in the authors list in galaxy.yml: https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/galaxy.yml


jilkarnas:
For variable pxe_stack_post_install_action, would it be better (e.g. more transparent/cleaner) to set default  value to 'reboot' and not "empty value" which reports push the definition of default value to reboot down the line in the j2 templates ?

oxedions:
Dear @jilkarnas 

Many thanks for this super PR ! :blush: 

Yes, I think it would be more transparent to make the default to "reboot", but please keep also the same default in the templates, because I could see a lot of users creating inventories with many values empty.

I also added a small fix for user-data file for Ubuntu, since preseed was only used for Ubuntu 18.04, now all other versions are using the user-data, which also contains a shutdown key.

oxedions (approved):
Dear @jilkarnas 

This is perfect ! Many many thanks ! :blush: 

Merging.

PR #1071, opened 2025-04-17
Fix: grafana - service management timeout

thiagocardozo:
A task that manages Grafana service might fail in timeout when using an external database.
This is due to extensive operations performed by Grafana before it returns a positive feedback.

Also fixed restart service handler variables.


ginomcevoy:
Looks good! Merging.

PR #1072, opened 2025-04-17
Docs: prometheus tls encryption

lmagdanello:
## Describe your changes

To enable TLS, you need to generate a certificate and private key. I've documented how the SysAdmin can use OpenSSL to quickly generate a self-signed certificate for use in Prometheus.

- [x] Make sure your name is present in the authors list in galaxy.yml: https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/galaxy.yml


ginomcevoy:
Got conflict on the CHANGELOG. Thiago set version to 3.2.5, I think it was the right decision for us.

PR #1092, opened 2025-04-30
Fix bluebanquise-diskless option 4 link nfs when node runs SELinux

ginomcevoy:
- Fix bluebanquise-diskless option 4 link nfs when node runs SELinux ,error were:

```
ERROR: command /usr/bin/grep -q '^smc:' /var/www/html/pxe/diskless/nfs_reference/rhel_9.5_aarch64-ref-standard/etc/passwd || /usr/sbin/useradd  -R /var/www/html/pxe/diskless/nfs_reference/rhel_9.5_aarch64-ref-standard/ -m -c 'SMC user' -d /opt/smc/home -s /bin/bash -p '$6$wDehqATpDnyU8il3$jGFLpIeN4F74xtdFAQfS7n3as9.FkONxpp4wNrNPYrzafdZW0GB3duYXAI95.AhFQEa.lr.3VFwoGL0dj9sDU0' smc failed :(
 Report:
 Error code: 9
 stdout: b'useradd: group smc exists - if you want to add this user to that group, use -g.\n'
```

This error can be fixed by creating "/sys/fs/selinux/enforce" in NFS image with contents "0".

- Second error appeared right after first fix
```
 ERROR: command /usr/bin/grep -q '^smc:' /var/www/html/pxe/diskless/nfs_reference/rhel_9.5_aarch64-ref-standard/etc/passwd || /usr/sbin/useradd  -R /var/www/html/pxe/diskless/nfs_reference/rhel_9.5_aarch64-ref-standard/ -m -c 'SMC user' -d /opt/smc/home -s /bin/bash -p '$6$wDehqATpDnyU8il3$jGFLpIeN4F74xtdFAQfS7n3as9.FkONxpp4wNrNPYrzafdZW0GB3duYXAI95.AhFQEa.lr.3VFwoGL0dj9sDU0' smc failed :(
 Report:
 Error code: 9
 stdout: b'useradd: group smc exists - if you want to add this user to that group, use -g.\n'
```

Fix now is to pass "-g" parameter to adduser.

- SMC ONLY: we are hard-coding the group to "smc" instead of a group with the same name as "sudo_user_home".
 Fix to master branch will  have to use "sudo_user_home".

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Format: <date> - <author (username or mail or both)> - [role] <change descriptio
 
 # 3.2.5
 
+* 4/30/25 - ginomcevoy - [pxe_stack] Fix issue with bluebanquise-diskless and host node running SELinux.
 * 4/30/25 - lmagdanello - [prometheus] Fix Prometheus Job template.
 * 4/29/25 - ginomcevoy - [podman] fix systemd for registry service to allow rc=2, use recommended Type=forking
 * 4/25/25 - thiagocardozo - [pxe_stack] Fixed when conditional check.
--- a/collections/infrastructure/roles/pxe_stack/files/bluebanquise-diskless
+++ b/collections/infrastructure/roles/pxe_stack/files/bluebanquise-diskless
@@ -99,6 +99,7 @@ import subprocess
 import crypt
 import string
 import sys
+import platform
 
 
 # Colors, from https://stackoverflow.com/questions/287871/how-to-print-colored-text-in-terminal-in-python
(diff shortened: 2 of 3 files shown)


ginomcevoy:
Note: "-g" option in useradd + --root fails when host node is running RHEL 8. The flag was included in original bluebanquise-diskless, but we removed it for SMC for to this reason.

Adding a commit so that the flag is skipped if host node is running RHEL 8, this fixes an issue and allows us to use RHEL 8 nodes for testing the tool.

ginomcevoy:
Tested OK on dev node and on VM with SELinux enforcing, merging.

PR #1095, opened 2025-05-01
Fix: nfs - Fixed unresolvable server name in mount

thiagocardozo:
If network is defined in nfs share, client will always add {server}-{network} to mount list.
This will work if server is a name as this entry will be added to /etc/hosts,
but not if it's an IP as it renders something like 10.0.0.1-net1.

This PR adds a check to verify if server name is an IP before appending network name.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Format: <date> - <author (username or mail or both)> - [role] <change descriptio
 
 # To 3.2.6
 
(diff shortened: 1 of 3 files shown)


ginomcevoy:
Tested OK in HA environment

oxedions (approved):
Dear @thiagocardozo
This is ok for me ! make sens. I just took time to check that ansible.utils.ipv4 is part of the basic ansible pip install, it seems ok :blush:

PR #1101, opened 2025-05-05
Fix: grafana: wrong firewall variable name

thiagocardozo:
There's a remaining ep_firewall variable in grafana role.


oxedions (approved):
Good catch !
Thank you 😊

PR #1108, opened 2025-05-15
Fix: pxe_stack: Dont recursively act on folder

thiagocardozo:
In pxe_stack theres a task that creates diskless nfs directories structure.
That's ok for when the role runs for the first time, but the task has a "recursive" property that will cause the module
to act on hundreds of thousands of files if a few dozen images are present. This is causing long delays and might even
break the images.

It seems that this task should create the folder and nothing else.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Format: <date> - <author (username or mail or both)> - [role] <change descriptio
 
 # To 3.2.7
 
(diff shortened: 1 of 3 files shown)


oxedions (changes requested):
Dear @thiagocardozo 

Thank you !
Yes it makes sens !
Could you just add an entry in the CHANGELOG ? 😊

oxedions (approved):
Super, thanks a lot ! :blush:

PR #1119, opened 2025-07-17
Improve node stability with configuring Slurm to keep some memory amount for the operating system

patrick-begou:
## Describe your changes
In Slurm, `mem_spec_limit` key is the amount of memory that Slurm must not use (minimum reservation for the OS).
When a user allocate a full node, some system instabilities can occur with OOM killer.
If in the node inventory "reserved_os_memory" (in MB) is specified, this limitation on usable amount of memory is introduced in slurm.conf 

This is a new feature to improve the minimal Slurm setup. It is running on my cluster after some node crashes with out-of-memory problems. Still in test (waiting for some production time to check if memory related instabilities are solved).

It will be finalized if Bluebanquise team agree with this improvement.

- [X ] Document introduced changes / variables in role README.md if any
- [ ] Make sure your name is present in the authors list in galaxy.yml: https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/galaxy.yml
- [ ] Update CHANGELOG file at repository root: https://github.com/bluebanquise/bluebanquise/blob/master/CHANGELOG

### DIFF
--- a/collections/infrastructure/roles/slurm/README.md
+++ b/collections/infrastructure/roles/slurm/README.md
@@ -241,6 +241,7 @@ hw_specs:
     sockets: 2
     threadspercore: 1
   memory: 63500
+  reserved_os_memory: 4096
   slurm_extra_nodes_parameters: Feature=XXXX Weight=YY
 ```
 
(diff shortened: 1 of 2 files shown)


patrick-begou:
Duplicate with slurm_extra_nodes_parameters

PR #1137, opened 2025-09-22
Fix: dhcp_server - bmc mac required

thiagocardozo:
A check in the subnet jinja2 template is forcing a BMC to have a MAC address defined,
causing the write to skip if for instance just dhcp_client_identifier is used.

Removed the check since it's already present in the write_host macro.


oxedions (changes requested):
Dear @thiagocardozo 

Thank you for this PR ! 😊 
I see a lot of files changes, I think your branche diverged. Could you check ?

PR #1153, opened 2025-11-14
[BUG] alertmanager: Ubuntu vars package is prometheus-alertmanager

fabienarchambault:
## Describe your changes
In Ubuntu noble (24.04.3 LTS), the package `alertmanager` does not exists. It is named `prometheus-alertmanager`.

All other packages are fine.

- [ ] Make sure your name is present in the authors list in galaxy.yml: https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/galaxy.yml
- [ ] Update CHANGELOG file at repository root: https://github.com/bluebanquise/bluebanquise/blob/master/CHANGELOG

### DIFF
--- a/collections/infrastructure/roles/prometheus/vars/Ubuntu.yml
+++ b/collections/infrastructure/roles/prometheus/vars/Ubuntu.yml
@@ -2,7 +2,7 @@
 prometheus_server_prometheus_packages_to_install:
   - prometheus
 prometheus_server_alertmanager_packages_to_install:
-  - alertmanager
+  - prometheus-alertmanager
 prometheus_server_ipmi_exporter_packages_to_install:
   - ipmi-exporter
   - freeipmi


oxedions:
That would be possible, but could be an issue later if RedHat chooses to provide alertmanager and use another naming convention than Debian/Ubuntu.

I was thinking doing it this way:

Rename packages to prometheus-xxxxxxx, and in the rpms mark as obsolete the xxxxxxx. So when installing prometheus-alertmanager, it would replace alertmanager. That is for rpm based systems.

```
Name:     prometheus-alertmanager     # <<<<<<<<<<<<<<<<<<< HERE
Summary:  prometheus-alertmanager  # <<<<<<<<<<<<<<<<<<< HERE
Release:  1%{?dist}
Version:  %{_software_version}
License:  apache-2.0
Group:    System Environment/Base
Source:   https://github.com/prometheus/alertmanager/releases/download/v%{_software_version}/alertmanager-%{_software_version}.linux-%{_software_architecture}.tar.gz
URL:      https://github.com/prometheus
Packager: Oxedions <oxedions@gmail.com>

Obsoletes: alertmanager  # <<<<<<<<<<<<<<<<<<< HERE
```

For Debian/Ubuntu, since I could find all Prometheus packages in Deb 12/13 and Ubuntu 22/24, I will no more distribute them. To allow users fluently upgrade, I propose to add a small temporary task in the role that would make sure packages [ "alertmanager", "node-exporter", ... ] are not present before installing OS packages. We can then remove this task in the future.
Real Ubuntu/Debian support is recent in BB, and there are in theory only few clusters not based on RHEL, so it should be ok.

Would that be ok on your side too ?

PR #1161, opened 2025-11-19
fix apt sources cleanup process 

dietergorecki:
In the end of ubuntu autoinst, repos are used during install are cleaned only in `/etc/apt/sources.list` which not used anymore. Sources are now located under `/etc/apt/sources.list.d/`
I kept it for compatibility purpose.


oxedions (approved):
Dear @dietergorecki 

Many thanks for this PR ! :blush: 

Makes sens ! Indeed, more and more files are in foobar.d/ folders, which is nice !

Merging :wink:

PR #1169, opened 2025-11-21
Enhance way to apply connection modifications

dietergorecki:
Hi,
Before this change I frequently needed to do `nmcli con up xxx` to properly apply connection changes, especially on bond setups when slave member where already partially configured.   
With this change it is automatically done at connection reload. I also preferred to use the decicated nmcli module instead of a shell command, probably more efficient.


oxedions:
Dear @dietergorecki 

Many thanks for this fix ! :blush: 

In case you tested it: do you know if it also covers gateway/routes update ? I discovered that when the gateway or routes are updated on the interface, a simple reload was not enough, I had to nmcli con down and nmcli con up to apply the changes.

PR #1171, opened 2025-11-24
Allow to define Ubuntu autoinstall ssh.allow-pw

btravouillon:
Sysadmins should be able to disable allow-pw during the deployment to enhance security.

## Describe your changes

Add a new variable `pxe_stack_ssh_allow_pw` which allows to define `ssh.allow-pw` in Ubuntu Autoinstall.
Defaults to `true` for backward-compatibility.


oxedions (changes requested):
Dear @btravouillon 

Thank you for this PR :blush: 

It makes perfect sens I agree. Would you accept to rename variable to `pxe_stack_ssh_allow_passwords` ? I will also add the feature to Debian and RHEL auto installation files with this same variable.

PR #1175, opened 2025-11-27
Feat: conman - use specific file for storing passwords

thiagocardozo:
Passwords stored in conman.conf are visible to anyone in the cluster with a simple ps:

`[root@host01 (HOST01) bluebanquise]# ps -edf | grep conman
conman    283979       1  0 19:45 ?        00:00:00 /usr/sbin/conmand -P /var/run/conman/conmand.pid conman    283980  283979  0 19:45 ?        00:00:00 /usr/bin/expect -f /usr/share/conman/exec/ipmitool.exp 10.1.0.214 USER PASSWORD`

By moving them to the conman.pwd file they'll be visible only if the user can open the file:

`[root@host01 (HOST01) bluebanquise]# ps -edf | grep conman
conman    283979       1  0 19:45 ?        00:00:00 /usr/sbin/conmand -P /var/run/conman/conmand.pid conman    283980  283979  0 19:45 ?        00:00:00 /usr/bin/expect -f /usr/share/conman/exec/ipmitool.exp 10.1.0.214 USER`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,7 +6,7 @@ Format: <date> - <author (username or mail or both)> - [role] <change descriptio
 
 * 11/28/25 - btravouillon - [pxe_stack] Ensure Ubuntu ssh.allow-pw is a boolean (#1178)
 * 11/28/25 - oxedions - [all] Update firewall to allow new bb_services_firewall_zone variable and switch all os on firewalld (#1162 and #1163)
-* 11/27/25 - thiagocardozo - [conman] Passwords in database file for increased security (#1174)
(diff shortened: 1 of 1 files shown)


collections/infrastructure/roles/conman/templates/conman.pswd.j2
```diff
@@ -0,0 +1,30 @@
+#jinja2: lstrip_blocks: "True"
```
oxedions:
```suggestion
#jinja2: lstrip_blocks: True
```

PR #1182, opened 2025-11-30
Ensure systemd-networkd is stopped in nic role

btravouillon:


oxedions:
Dear @btravouillon 

Thanks a lot for the feedback !

I will merge it. :slightly_smiling_face: 

On another cluster, I saw that time to time (I think its when I update system), systemd-networkd was back alive again (probably triggered by a post-script in a package).
Making sure that it is stopped when executing the role is I think a secure thing. And if it worked with your test, then it means it is reproducible.

Issue #10, opened 2019-09-12
Contribution guidelines


bouriquet15:
Hi,

It could be great to have some contribution guidelines (contrib.md, role_skeleton for ansible-galaxy)

Skeleton example (you just need to provide a role directory template somewhere in this repo) : https://github.com/cjsteel/galaxy-role-skeleton
Usage : ansible-galaxy init --role-skeleton=<skeleton_dir> <role_name>

johnnykeats:
Hi!

Thanks for the feedback.
I tried to create a contribution file. Hope this is what you are expecting. I will try to "make it live" as we encounter new situations.

Regarding Skeleton for Ansible Galaxy, this is is something I am not familiar with, I will let Oxedions work on this :blush:

oxedions:
Hello,

Thank you JK for the contrib file ;-)

So for what I understand, we should provide a skeleton of our roles for an usage in the Ansible Galaxy. Could be a good opportunity to comply with what is already done in the Ansible world.

The tree generated with the ansible-galaxy command contains much more folders than what we have today, and generated a readme in MD, while we rely on RST for the documentation. I need to check this.

@bouriquet15 , do you know if in this skeleton I need to force things like the license ? Sorry, I am not very used to Ansible Galaxy :-)

Best regards

Ox

oxedions:
No answers, so closing for now.

PR #1188, opened 2025-12-02
update kernel_config role to add support for ubuntu (#10)

dietergorecki:
By default only Redhat is supported (grubby is not in ubuntu repositories). Here we use the standard update-grub method.

### DIFF
--- a/collections/infrastructure/roles/kernel_config/defaults/main.yml
+++ b/collections/infrastructure/roles/kernel_config/defaults/main.yml
@@ -1,2 +1,5 @@
 ---
 kernel_config_sysctl_reload: true
+os_kernel_parameters: ""
+hw_kernel_parameters: ""
+hw_console: ""
--- a/collections/infrastructure/roles/kernel_config/handlers/main.yml
+++ b/collections/infrastructure/roles/kernel_config/handlers/main.yml
@@ -1,0 +1,3 @@
+---
+- name: command <|> Update GRUB settings
+  ansible.builtin.command: update-grub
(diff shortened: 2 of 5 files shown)


oxedions (commented):
Dear @dietergorecki 

Thank you for this PR ! :blush: 

I provide grubby in the BlueBanquise repositories, to uniformize roles:
https://bluebanquise.com/repository/releases/latest/u24/x86_64/bluebanquise/pool/main/g/grubby/

You can use file https://bluebanquise.com/repository/releases/latest/u24/x86_64/bluebanquise/bluebanquise.list to enable BB repos.

Would that solve the issue or do you think we still need to add the grub2 method for Ubuntu and Debian systems ? I dont really have a preferred way to be honest.

dietergorecki:
OK I haven't thought about checking grubby on your repo.  
I don't know how grubby handles kernel settings on ubuntu, but I guess my method is closer to the way people are to do it under ubuntu. I would still understand if you prefer to stick with one method for multiple OS as it makes less code to maintain.  
Decision is up to you, I wouldn't take offense if you don't merge it :)

oxedions:
Dear @dietergorecki 

My apologies for the delay. Last days were super heavy, I couldnt have free time to work on BB.

I did some investigations, and I must agree with you, its definitely the Debian/Ubuntu way, and I always prefer to bind to what is recommended for each OS. I will stop building grubby for Debian and Ubuntu once this new code is release.

So ok for me, merging ! :blush:

PR #1203, opened 2025-12-23
Enable logging support in sudoers (#40)

dietergorecki:
Hi,  
To be able to log all sudo activities, use the log_output option. It is disabled for the user used with ansible to avoid over logging.


oxedions (dismissed):
Dear @dietergorecki 

Super useful indeed ! Thank you for this PR :blush: 

Merging !

oxedions (approved):

PR #42, opened 2019-12-03
Update bootset.py

bouriquet:

PR #1204, opened 2025-12-23
Copy hosts file, don't generate (#42)

dietergorecki:
Introduce an option (disabled by default) to copy instead of generate hosts file on all hosts. Copy is unefficient: too long and unneeded.
On nodes with management roles, hosts file is always generated.

### DIFF
--- a/collections/infrastructure/roles/hosts_file/defaults/main.yml
+++ b/collections/infrastructure/roles/hosts_file/defaults/main.yml
@@ -8,3 +8,6 @@ hosts_file_fqdn_first: false
 # If set to true, hostname, alias, nic-network and nic-alias will be written info hosts file
 # If set to false, only hostname and alias will be written into hosts file
 hosts_file_enable_extended_names: true
+
+# Copy hosts file instead of generate it for all hosts
+hosts_file_copy: false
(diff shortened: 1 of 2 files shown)


oxedions:
Dear @dietergorecki 

Thank you for this PR :blush: 

I am aware of this issue. I was asked to scale to 3000 hosts for next BB version, and this hosts file generation is a pain.

Your code is ok, and since its optional we can merge it.
I wanted to propose this, would that be also an option for you too ? (asking this because this prevent the need to first generate the hosts file on the ansible controller and then have to push it, so 2 steps).

1. Template is now stored into a variable (in vars/main.yml)
2. In the task, we render the template once (Atomic operation, so only calculated once, whatever the number of hosts executed in parallel), and it is stored into a register.
3. Next step is just a write of the content of the register to /etc/hosts for each hosts.

Doing this, you only calculate the content once instead of having to recalculate the same things for each hosts of the range of hosts.
I wanted to do the same for slurm.conf file, because this is the exact same issue: big file to be written on all worker hosts.

oxedions:
Hey @dietergorecki :blush: 

Any news on this topic ?

PR #1219, opened 2026-02-24
SMC: Fix - credentials host error and add REDFISH support

thiagocardozo:
* Added missing REDFISH matching block
* Removed initial _hw_board_authentication_ var definition check
* Use _credentials_host_ var instead of _host_ because _hw_board_authentication_ might be defined in BMCs


giacomo-mcevoy:
Tested all combinations OK:
- IPMI in compute equipment group
- REDFISH in compute equipment group
- IPMI in bmc equipment group
- REDFISH in bmc equipment group

OK for me to merge. Don't forget to fix master :)

PR #1227, opened 2026-03-20
BRSMC-2323 - Replace wrong variables in template

thiagocardozo:
Template was using "prometheus_server_alertmanager_*" instead of "prometheus_server_prometheus_*"
for setting prometheus server credentials.


ginomcevoy:
Hello @thiagocardozo , does this mean that the communication between alertmanager and prometheus was failing? It seems to be that it would have triggered an error in integration, unless they are not testing this part.

thiagocardozo:
It means the communication will fail if the credentials are different.
If they are equal no one will notice.

ginomcevoy (approved):
OK I now understand the ticket, thanks for the info. Approving.

PR #1265, opened 2026-07-26
[dns_server] Fix dns server override mechanism

santos-lucas:
2 related changes on this PR:

1. Removed the condition `dns_server_zoneconf.changed` to generate the dns_server_serial and to generate the override file - otherwise the step to generate will be skipped and changing only `dns_server_overrides` will have no effect when running the role.

2. Fixed broken variables in [collections/infrastructure/roles/dns_server/templates/override.j2](/bluebanquise/bluebanquise/blob/master/collections/infrastructure/roles/dns_server/templates/override.j2)

Please use this checklist to help me fasten the review.

- [x] Document introduced changes in main documentation (see documentation folder)
- [x] Make sure your name is present in the authors list in galaxy.yml: https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/galaxy.yml (optional, up to you)

:warning: Ensure your PR does not break static code analysis (see automatic checks).

### DIFF
--- a/collections/infrastructure/roles/dns_server/tasks/main.yml
+++ b/collections/infrastructure/roles/dns_server/tasks/main.yml
@@ -113,7 +113,6 @@
 - name: set_fact <|> Set SOA Serial for reverse zone
   ansible.builtin.set_fact:
     dns_server_serial: "{{ lookup('pipe', 'date +%s') }}"
-  when: dns_server_zoneconf.changed  # noqa no-handler
   tags:
     - template
     - internal
@@ -162,7 +161,7 @@
     group: root
     mode: 0644
   notify: service <|> Restart dns services
-  when: dns_overrides is defined and dns_server_zoneconf.changed  # noqa no-handler
+  when: dns_server_overrides is defined # noqa no-handler
   tags:
     - template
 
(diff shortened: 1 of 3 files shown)


oxedions (approved):
Dear @santos-lucas 

Thank you for this PR ! 😊 

It makes perfect sens, please just check my comment.

collections/infrastructure/roles/dns_server/tasks/main.yml
```diff
@@ -113,7 +113,6 @@
 - name: set_fact <|> Set SOA Serial for reverse zone
   ansible.builtin.set_fact:
     dns_server_serial: "{{ lookup('pipe', 'date +%s') }}"
-  when: dns_server_zoneconf.changed  # noqa no-handler
```
oxedions, line 116: I don't understand this change however. It will make this task always execute even when not needed.

santos-lucas: Yeah, I agree with you, I was wondering about that also, maybe we can improve it, but at this point the issue is: if you just change the "dns_server_overrides" for example, the zoneconf is not changed, and overrides.j2 template cant be rendered because it misses the `dns_server_serial`

oxedions: Sorry for the delay 😊 

Ok, makes sens. Lets go this way then.

oxedions (approved):

